PR #3, opened 2023-07-14
Initial set of architectural decisions

jazairi:

We want to record the decisions we make about this system.

#### Relevant ticket(s):

N/A.

This adds a few initial ADRs:

* The decision to use ADRs.
* The decision to use Rails.
* The decision to implement a REST API.

While these ADRs are pending approval, I will likely begin some work that assumes they have been approved. I will not merge any such PRs until we have had the opportunity to discuss and approve the decisions.

### DIFF
--- a/docs/architecture-decisions/0001-record-architecture-decisions.md
+++ b/docs/architecture-decisions/0001-record-architecture-decisions.md
@@ -1,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2023-07-14
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
+
+## Consequences
+
+See Michael Nygard's article, linked above. For a lightweight ADR toolset, see Nat Pryce's [adr-tools](https://github.com/npryce/adr-tools).
--- a/docs/architecture-decisions/0002-use-rails.md
+++ b/docs/architecture-decisions/0002-use-rails.md
@@ -1,0 +1,21 @@
+# 2. Use Rails
+
+Date: 2023-07-14
+
+## Status
+
+Accepted
+
+## Context
+
+We need to choose a language and framework in which to write this application.
+
+## Decision
+
+We will use Ruby on Rails.
+
+## Consequences
+
+Using Rails leverages preexisting expertise in EngX and will allow us to build out TACOS more rapidly. At this point,
+we are not aware of any functionality this tool will need that Rails cannot provide. However, EngX is the only team in
+DLS that uses Rails, so we are committing to sole ownership and maintenance of the application.
--- a/docs/architecture-decisions/0003-implement-as-a-graphql-api.md
+++ b/docs/architecture-decisions/0003-implement-as-a-graphql-api.md
@@ -1,0 +1,85 @@
+# 3. Implement as a GraphQL API
+
+Date: 2023-07-14
+
+## Status
+
+Accepted
+
+## Context
+
+In early design exercises, we envisioned TACOS more in terms of the staff workflow of analyzing search terms. In the
+longer term, we would also want the tool to be able to evaluate search terms algorithmically.
+
+In either case, the tool should be able to return its evaluation of a search term to a search interface, so that
+interface can better detect user intent. An API will allow for any discovery interface to programmatically request data
+about search terms, and for TACOS to return that data efficiently and in a structured format.
+
+The basic flow might look like this:
+
+1. A user pastes a citation into the Bento input field and executes a search.
+2. Bento asks TACOS (possibly via ActiveRecord callbacks) to analyze the search term.
+3. TACOS evaluates the search term and hopefully identifies it as a citation.
+4. TACOS returns its evaluation to Bento.
+5. In the search results, Bento shares its best guess as to what the user was searching for. In this case, it would be
+a link to the article, but it might also be a database or journal, a library service offering, or contact information
+of a subject matter expert in the libraries.
+
+```mermaid
+flowchart LR
+  bento
+  tacosapi
+  analyze
+  returndata
+  ui
+
+  bento[1. Bento Search] -- API Request --> tacosapi[2. TACOS API] --> analyze[3. analyze with various algorithms] --> tacosapi 
+  
+  tacosapi -- API Response --> returndata[4. returnStucturedData] --> ui[5. Use data in UI]
+
+```
+
+This pattern would also replace the 'fact' feature built into the TIMDEX UI. In other words, TACOS would be responsible
+for matching DOIs, ISBNS, ISSNs, and other identifiers that could resolve known-item searches whereas consuming
+applications such as TIMDEX UI would be responsible for presenting the information in a useful way.
+
+### GraphQL vs REST
+
+We have familiarity with both GraphQL and REST, and either option is a viable choice for this API. We are ignoring other
+specifications, such as JSON:API, which are not not worth pursuing as they would slow down initial development and are
+unlikely to add significant value.
+
+#### REST
+
+For a single endpoint API like we are imagining, the simplicity of REST is a nice benefit. We are imagining minimal parameters and a single structured JSON data response and REST is very clean to use for this purpose.
+
+REST is a ubiquitous architectural style that is more established than GraphQL. If we plan to make TACOS something that
+other organizations can fork and customize for their own needs, then REST will lower the barrier to engage with the
+codebase.
+
+If we choose REST, we should assume we'll need to include a breaking version at some point in the future and start our routes versioned (i.e. `/api/v1?searchterm?=` rather than just `/?searchterm=`).
+
+#### GraphQL
+
+GraphQL requires a bit more care to implement from a consumption perspective as it is likely we'd break each portion of the structured JSON data into an optional returned value so consumers would need to explicitly ask for the data they plan to use. This can also be a benefit as we'd only send consumers the data they need, so while there is friction it is fairly small once consumers embrace how GraphQL works.
+
+GraphQL would allow us to have an interactive dashboard, GraphiQL, similar to what we have for TIMDEX to allow for an easy way to see responses from the API in a simple, self documenting web UI.
+
+While we don't anticipate a need for this so early on, GraphQL's use of deprecation instead of versioning could come
+in handy if this API is long lived. We were able to leverage this well in TIMDEX.
+
+Setting up GraphQL enpoints intially would be a bit more work than REST, but we have experience on our team doing this so it is not a blocker.
+
+The ability to return individual fields in GraphQL could result in significant performance gains. Some algorithms may
+have a higher cost than others, such as those that must call an external API for additonal information about a term. If
+we run these algorithms selectively based on what a consumer requests, we could substantially improve the API's
+performance.
+
+## Decision
+
+We will implement the functionality to return search term evalutions as a GraphQL API.
+
+## Consequences
+
+By making this decision, we are forgoing the added simplicity and familiarity that REST provides. However, for this use
+case, the value added by GraphQL's self-documentation and selective returning of fields outweighs these advantages.
--- a/docs/architecture-decisions/0004-monolithic-application.md
+++ b/docs/architecture-decisions/0004-monolithic-application.md
@@ -1,0 +1,55 @@
+# 4. Monolithic application
+
+Date: 2023-07-20
+
+## Status
+
+Accepted
+
+## Context
+
+There are multiple ways to architect the application, or applications, to provide the type of functionality we are planning.
+
+### Monolithic Architecture
+
+One option is to build a monolithic Rails application that provides all of the planned functionality: an API to capture search terms and return structured data, the ability to analyze the search terms to determine what the structured data should be (i.e. is there a DOI in this term and if so we are confident it is for a specific item), as well as expert workflows to allow manual categorization of search terms.
+
+### Services Architecture
+
+Another option is to isolate each piece of functionality into separate services. Even within this approach, there are different ways to slice up the services to create a balance of additional application communication complexity with the benefit of single purpose applications.
+
+#### Workflow Application and Categorization Application
+
+This separates the functionality into two distinct applications.
+
+##### Workflow Application
+
+The Workflow Application handles staff user authentication and presents users with individual search terms that is must request from the Categorization Application. Users can write back to the Categorization Application via API when a category to assign a term to a Category. The Workflow Application focuses on authentication, presenting data from an API, and posting data to an API. It is likely users would also want to see analytics about searchterms so the Categorization Application would need to expose that via API as well as the Workflow Application stores no data at all.
+
+##### Categorization Application
+
+The Categorization Application exposes multiple APIs and has no user interface that is not an API. The API must restrict write access to APIs that can write categorizations, but the categorization lookup API is likely unrestricted (even though it would store the incoming searchterm in most cases). Whether we require authentication for applications to be able to lookup data is out of scope of this ADR, but will need to be considered.
+
+Lookup API: accepts at minimum a searchterm and a string representing the calling application. Returns structured data with everythig we know about the provided search term.
+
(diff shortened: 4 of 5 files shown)


docs/architecture-decisions/0002-use-rails.md
```diff
@@ -0,0 +1,21 @@
...
+We need to choose a language and framework in which to write this application.
+
+## Decision
+
+We will use Ruby on Rails.
+
+## Consequences
+
+Using Rails leverages preexisting expertise in EngX and will allow us to build out TACOS more rapidly. At this point,
+wwe are not aware of any functionality this tool will need that Rails cannot provide. However, EngX is the only team in
```
JPrevost, line 20: typo in first word on this line

JPrevost (commented):
This looks great, thanks for writing it up.

I want to think a bit more on whether there is value to use GraphQL vs REST before I sign off. I may add a section on the pros and cons of each for us to consider to this document and then we can discuss and make our decision.

docs/architecture-decisions/0003-implement-as-a-rest-api.md
```diff
@@ -25,6 +25,24 @@ of a subject matter expert in the libraries.
...
+
+For a single endpoint API like we are imagining, the simplicity of REST is a nice benefit. We are imagining a minimal parameters and a single structured JSON data response and REST is very clean to use for this purpose.
+
+If we choose REST, we should assume we'll need to include a breaking version at some point in the future and start our routes versioned (i.e. `/api/v1?searchterm?=` rather than just `/?searchterm=`).
+
+### GraphQL
+
+GraphQL requires a bit more care to implement from a consumption perspective as it is likely we'd break each portion of the structured JSON data into an optional returned value so consumers would need to explicitly ask for the data they plan to use. This can also be a benefit as we'd only send consumers the data they need, so while there is friction it is fairly small once consumers embrace how GraphQL works.
+
+GraphQL would allow us to have an interactive dashboard, GraphiQL, similar to what we have for TIMDEX to allow for an easy way to see responses from the API in a simple, self documenting web UI.
```
jazairi, line 40: GraphiQL is definitely a compelling reason to use GraphQL. Hm...

docs/architecture-decisions/0003-implement-as-a-rest-api.md
```diff
@@ -25,6 +25,24 @@ of a subject matter expert in the libraries.
...
+
+While we don't anticipate a need for this so early on, GraqphQL's deprecation instead of versioning could come in handy if this API is long lived. We were able to leverage this well in TIMDEX.
```
jazairi, line 42: Typo: `GraphQL's deprecation`

docs/architecture-decisions/0003-implement-as-a-rest-api.md
JPrevost: Yeah, the self documentation and GraphiQL is genuinely why I started second guessing REST.

docs/architecture-decisions/0003-implement-as-a-rest-api.md
JPrevost: I read this 10 times before I saw what the issue was lol. Thanks!

.gitignore
```diff
@@ -37,6 +37,7 @@ build-iPhoneSimulator/
 /_yardoc/
 /doc/
 /rdoc/
+.adr-dir
```
JPrevost, line 40: We actually check this in to the repo as it defines which folder to generate ADRs in. If we don't all have the same config in this file, new ADRs will end up in weird places.

jazairi: Will do. I excluded it because I assumed not everyone was using `adr-tools`, but it makes sense to include it for those that do.

docs/architecture-decisions/0003-implement-as-a-rest-api.md
```diff
@@ -0,0 +1,60 @@
...
+
+Setting up GraphQL enpoints intially would be a bit more work than REST, but we have experience on our team doing this so it is not a blocker.
+
+## Decision
+
+The functionality to return search term evalutions will be implemented as a REST API.
```
matt-bernhardt, line 52: At the moment I'm not sure that I can come to a position on this ADR. In gradient of agreement terms, I'm not going to block either decision - but I also can't get to a place where I support either choice without further discussion.

Some of the considerations about this choice, for me, are:
- Standardization with our other APIs (where we just retired REST in Timdex API, and am I'm not sure what other APIs we're either maintaining or asking our community to use).
- Whether our architectural principles discuss this question (the latest draft I see doesn't mention either)
- Whether the various uses of this system might need the flexibility offered by GraphQL (I don't have a clear picture of what those uses are right now, so I'm having a hard time getting a meaningful signal here).

Bottom line for me is that more discussion is needed.

docs/architecture-decisions/0004-monolithic-application.md
```diff
@@ -0,0 +1,49 @@
...
+
+- GET: provides either a single term that needs categorization or possible a few terms. It is possible this may return objects or just raw terms depending on the needs of the Workflow Application.
+
+- POST: writes back a category for the provided searchterm. It is possible this will be an object or array depending on what the Workflow Application requires.
+
+Analytics API: likely a series of endpoints or a single endpoint with multiple features that will allow accept parameters and return appropriate analytics. The specific have not yet been determined.
+
+## Decision
+
+[TBD]
```
matt-bernhardt, line 45: My opinion should be understood as a naive starting point at the beginning of a long term effort - but _for the moment_ I prefer a monolithic application.

My thinking for this goes in the face of our stated architecture principles (which at a high level prefer decoupled services), but I think this is acceptable because I don't yet have a clear picture of where the appropriate decoupling would happen. The usual warnings about premature optimization are persuasive for me in this moment.

In the long term, I suspect that everything we're talking about in this repository should not live together; by starting with a monolithic approach, then, we're inviting a later refactoring to accommodate a split. That is preferable, however, to realizing later that we need to combine functions from separate applications, potentially as N components are refactored to be a differently-sized set of N components.

matt-bernhardt (commented):
At the moment, I'm definitely in favor of merging the first two ADRs as is, and continuing discussion of the second two ADRs.

In terms of my actual positions on those last two ADRs, right now I'm very neutral about which API format to use, and in favor of starting this project off on a monolithic basis - but expecting that we'll end up splitting the app up down the road as we more fully implement its details.

docs/architecture-decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2023-07-14
+
+## Status
+
+Proposed
```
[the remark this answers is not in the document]
jazairi: My plan is to change them to "Approved" when we get three 👍's. (If I were to do this over again, I would have made each of these a separate PR so we approve them individually.)

I'm sure @JPrevost is also 👍 on this, but I'll wait for explicit approval before making the change.

docs/architecture-decisions/0002-use-rails.md
```diff
@@ -0,0 +1,21 @@
...
+
+Proposed
+
+## Context
+
+We need to choose a language and framework in which to write this application.
+
+## Decision
+
+We will use Ruby on Rails.
```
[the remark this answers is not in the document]
jazairi: 👍 for me as well. As with 0001, I assume @JPrevost is also 👍, but I'll wait for him to chime in before changing status.

docs/architecture-decisions/0003-implement-as-a-rest-api.md
jazairi: I believe the phrasing has not changed yet. I'm pushing up a commit to fix the noted typos now.

docs/architecture-decisions/0004-monolithic-application.md
jazairi: I agree with everything you've said here, but I'd like to manage expectations about the possibility of a future refactor. The likelihood that we would commit resources to that feels low to me.

docs/architecture-decisions/0003-implement-as-a-rest-api.md
jazairi: In gradients of agreement terms, I am probably at "support with reservations" or "agreement with a minor point of contention." I agree with your last bullet point questioning whether GraphQL's flexibility adds value to this use case. I would also suggest that REST's ubiquity could be advantageous to encourage engagement with the codebase, but I'm not sure if that is important in this instance.

Ultimately, neither of these concerns rise to a blocking level, particularly given GraphQL's built-in documentation features and the fact that TIMDEX is now fully GraphQL. API standardization is an interesting question, though, and I'd be curious if we're maintaining other REST APIs that I'm unware of.

JPrevost: Interesting. Do we envision non-MITL systems interacting with this API? If no, I don't think REST's ubiquity matters. I can see both sides. This is an internal tool we need to evaluate our understanding of search terms and hopefully return some interesting/useful bits of data about the terms... but what if it was open? Maybe we only send stuff from MITL to our MITL experts for categorization workflows but allow anyone to use the lookup portion of the API once we have some useful algorithms in place?

docs/architecture-decisions/0001-record-architecture-decisions.md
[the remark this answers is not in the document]
JPrevost: Ah, yeah. I think I usually leave the ADRs as "approved" when they are in the PR with the assumption it makes them easier to merge once we actually approve them. I'm 👍🏻 on ADR 1 and 2 and they can be merged once we update them to approved. I'm not sure how best to do that without losing any of the dialog here so it may be best to just leave them unmerged even though we agree with them and continue working on 3 and 4 and aim towards merging them all in the next week or so.

docs/architecture-decisions/0004-monolithic-application.md
JPrevost: If TACOS becomes useful, then considering a refactor to provide specific new features more efficiently will happen if the pain to add them to a monolith is high enough. For pathfinder work like this, I think simpler is often better and I am leaning towards a monolithic rails app that handles the categorization workflows as well as the algorithmic categorization. Having one app do both tasks has the added benefit of being able to have staff validate the algorithmic guesses as well.

The main downside is if we want to extract the algorithmic pieces into it's own API in the future, but that may be solvable even in a monolithic "feeling" application by moving them to their own Gems to be included in multiple places without reimplementation.

I guess in short, I don't see the value in moving staff categorization into a separate app from algorithmic detection. It either adds a much more complicated set of APIs that are likely to only ever be used between these two sets of features, or we cheat and have two apps share a single database (which is worse).

I'm still open to what features we gain by breaking up the monolith at this early stage, but if we can't say specifics I am not likely to be convinced.

docs/architecture-decisions/0002-use-rails.md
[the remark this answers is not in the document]
JPrevost: 👍🏻 for ADR 0002

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
```diff
@@ -0,0 +1,81 @@
...
+  ui
+
+  bento[1. Bento Search] -- API Request --> tacosapi[2. TACOS API] --> analyze[3. analyze with various algorithms] --> tacosapi 
+  
+  tacosapi -- API Response --> returndata[4. returnStucturedData] --> ui[5. Use data in UI]
+
+```
+
+This pattern would also replace the 'fact' feature built into the TIMDEX UI. In other words, TACOS would be responsible
+for matching DOIs, ISBNS, ISSNs, and other identifiers that could resolve known-item searches.
```
JPrevost, line 40: I'm going to add a sentence here to clarify the responsibilities of TACOS vs UIs that use it: TIMDEX UI would be responsible for deciding what to do when TACOS returns structured data that includes information about a DOI, etc

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
```diff
@@ -0,0 +1,81 @@
...
+Setting up GraphQL enpoints intially would be a bit more work than REST, but we have experience on our team doing this so it is not a blocker.
+
+The ability to return individual fields in GraphQL could result in significant performance gains. Some algorithms may
+have a higher cost than others, such as those that must call an external API for additonal information about a term. If
+we run these algorithms selectively based on what a consumer requests, we could substantially improve the API's
+performance.
+
+## Decision
+
+We will implement the functionality to return search term evalutions as a GraphQL API. An API will allow for any
```
JPrevost, line 72: I think a fair bit of this might be better in the consequences section.

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
```diff
@@ -0,0 +1,81 @@
...
+discovery interface to programmatically request data about search terms, and for TACOS to return that data efficiently and in a structured format.
+
+We have familiarity with both GraphQL and REST, and either option is viable. However, for this use case, the value added
+by GraphQL's self-documentation and selective returning of fields outweights REST's simplicity and greater ubiquity.
+
+## Consequences
+
+We are ignoring other specifications, such as JSON:API, which are not not worth pursuing as they would slow down initial
```
JPrevost, line 80: This probably belongs in the context section.

docs/architecture-decisions/0004-monolithic-application.md
```diff
@@ -0,0 +1,53 @@
+# 4. Monolithic application
```
JPrevost, line 1: 👍🏻 on this ADR

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
```diff
@@ -0,0 +1,81 @@
+# 3. Implement as a GraphQL API
```
JPrevost, line 1: 👍🏻 on the ADR conceptually, I have a few wording suggestions I'd like to implement (mostly moving a few bits around) but didn't want to do that yet in case others preferred it as-is (I wouldn't block moving forward exactly as written so my 👍🏻 is real)

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
jazairi: I disagree; I think it's more contextual. I'm about to push up a rewrite that moves the `GraphQL vs REST` section beneath `Context`, and puts everything in the `Decision` and `Consequences` section up there. Let me know if that makes more sense to you.

docs/architecture-decisions/0004-monolithic-application.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Decision
+
+We will build this application initially as a monolithic application.
+
+## Consequences
+
+We will approach this as a pathfinder project, where we learn as we go and update our path and our plans dynamically as new information is uncovered.
+
+In order to maximize our efficiency, we won't try to predict the "best" set of applications that will underpin this type of system before we start, and instead will reflect as we go and design the monolothic application modularly so that if a portion feels like it should spin off into its own external service, we will be able to interact via the same internal object in the monolith and no other objects will have to know anything has changed.
```
jazairi, line 51: This is not really something we've done yet, at least that I'm aware of. I think it addresses my concern about the low likelihood of major post-launch refactors, but it may be worth noting that this will be a new practice.

docs/architecture-decisions/0004-monolithic-application.md
jazairi: 👍 On this ADR. (FYI, I pushed a few commits fixing typos; none of them changed any wording.)

matt-bernhardt: 👍 on this ADR

docs/architecture-decisions/0003-implement-as-a-graphql-api.md
matt-bernhardt: 👍 on this ADR

matt-bernhardt (approved):

docs/architecture-decisions/0004-monolithic-application.md
JPrevost: I concur. I'll add a sentence to reflect that.

JPrevost (approved):

PR #4, opened 2023-10-12
Updates TACOS Class diagram

JPrevost:

* We have had discussion and made decisions that need to be reflected in the Class Diagram

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ENGX-231

* Batch Class is removed (and batch_id is removed from Search)
* Rename Search to SearchEvent
* Add source to SearchEvent
* Move search_phrase to Term
* Remove list_unique_terms() from Term
* Relationship between Term and SearchEvent is 1:many

### DIFF
--- a/docs/reference/classes.md
+++ b/docs/reference/classes.md
@@ -14,7 +14,6 @@ classDiagram
     Term: id
     Term: +String phrase
     Term: calculate_certainty(term)
-    Term: list_searches()
     Term: list_unique_terms_with_counts()
     Term: uncategorized_term()
     Term: categorized_term()
--- a/docs/reference/classes.md
+++ b/docs/reference/classes.md
@@ -3,31 +3,27 @@ classDiagram
   direction TB
   
   AdminUser --> User : Is a Type of
-  Term --> Search : Represents a common phrase in a set of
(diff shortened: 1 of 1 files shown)


docs/reference/classes.md
```diff
@@ -3,31 +3,27 @@ classDiagram
...
-  Batch --> Search : Loads and Processes
   Categorization --> Term : Includes a
   Categorization --> Category : Includes a
 
   class Term
-    Term
     Term: id
+    Term: +String phrase
     Term: calculate_certainty(term)
     Term: list_searches()
```
jazairi, line 17: Could we change this to `list_search_events()`, now that we've changed the name of that class? (Or just remove it; as you noted in our meeting this morning, it's effectively a convenience method.)

jazairi (changes requested):
See my comment about changing the `list_searches()` method name. Otherwise, I think this is ready to go. Thanks for making these updates!

docs/reference/classes.md
JPrevost: I chose to remove it (at least for now) from the Class.

jazairi (approved):

PR #5, opened 2023-10-12
Rails71

JPrevost:
Rails 7.1.1 initialized in the existing repo via: 
`rails new .`

The value to add to `config/master.key` is located in LastPass for this application.

* https://mitlibraries.atlassian.net/browse/ENGX-235

As we have no CI yet, reviewers may want to:

- create `config/master.key` locally and add the value from our LastPass
- `bundle install`
- `bin/rails test` to confirm no errors
- `bin/rails s` and then access the running app to see the Rails welcome screen


jazairi (commented):
I think this is all good, but I'm running into something weird locally. When I run `bundle install`, it adds ARM as a platform to Gemfile.lock and installs ARM versions of Nokogiri and SQLite. Is that something you've encountered?

JPrevost:
@jazairi my Mac is an older Intel model. I'm assuming what you are seeing is expected if you have an M1/M2 Mac. Can you add a commit to this branch with your updated lock file?

jazairi:
Ah, I didn't realize you had an Intel Mac. That explains it.

jazairi (approved):

PR #6, opened 2023-10-19
Adds Searchevent and term

JPrevost:

* this implements the first two models and will enable the storage of
  the data necessary to get started with this application

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ENGX-232
https://mitlibraries.atlassian.net/browse/ENGX-233

* Rails

- some of our typical developer gems, rubocop and annotate were added
- rubocop was used to autoformat everything in the `/app` and `/test`
  directories
- adds our standard CI setup
- adds dependabot weekly checks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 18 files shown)


jazairi (approved):
This is consistent with my understanding of the most recent version of the class diagram. I suspect we'll want an `initialize` method on the Term class, but we can add that later if it makes sense when we start to do more with these models.

Thanks for doing all the Rails and CI setup, too; in retrospect, that probably should've been a separate ticket!

PR #11, opened 2023-11-01
Initial GraphQL implementation

jazairi:

We need a GraphQL interface to record SearchEvents from TACOS
consumers.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ENGX-234

* Adds the GraphQL gem and generates a new install.
* Adds a SearchEvent type and corresponding query field that
returns information about the search event and creates a new
SearchEvent (and Term, if applicable).

* Removed a couple of generated GraphQL test fields.
* Miscellaneous rubocop fixes.

### DIFF
--- a/app/graphql/types/search_event_type.rb
+++ b/app/graphql/types/search_event_type.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class SearchEventType < Types::BaseObject
+    field :id, ID, null: false
+    field :term_id, Integer
+    field :source, String
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,8 @@
 Rails.application.routes.draw do
+  if Rails.env.development?
+    mount GraphiQL::Rails::Engine, at: "/graphiql", graphql_path: "/graphql"
+  end
+  post "/graphql", to: "graphql#execute"
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
   # Reveal health status on /up that returns 200 if the app boots with no exceptions, otherwise 500.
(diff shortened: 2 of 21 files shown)


jazairi:
To address the coverage issues seems like we would be testing the gem. I'm curious about differing perspectives on this, and whether additional coverage is needed outside of what's being reported.

app/graphql/types/search_event_type.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class SearchEventType < Types::BaseObject
```
JPrevost, line 4: Thought for the future: I'm waffly (and thus not requesting a change) as to whether it might be good to include the `Term.phrase` as part of this externally facing object. I know from the data perspective, it is a relation but from the API perspective we may want this to feel like a single object. Hmmm.

config/routes.rb
```diff
@@ -1,4 +1,8 @@
 Rails.application.routes.draw do
+  if Rails.env.development?
+    mount GraphiQL::Rails::Engine, at: "/graphiql", graphql_path: "/graphql"
```
JPrevost, line 3: In TIMDEX we've decided to implement GraphiQL as an HTML page instead of via the Gem. I don't remember why, so this isn't a request for a change but just something we should keep our eye on to know whether we should update TIMDEX or TACOS to match the other.

https://github.com/MITLibraries/timdex/blob/main/public/playground.html

If we can configure the Gem version, it may be better as it would fit into our dependency update patterns more easily than the standalone HTML does. 🤷🏻

JPrevost (approved):

JPrevost:
@jazairi I added a commit to address the mutation warning. I am not (currently at least) concerned with the code coverage as that is all generated code as you noted.

app/graphql/types/search_event_type.rb
jazairi: I had the same thought and, being similarly ambivalent, stuck to what is contained in the `SearchEvent` model for now. My expectation is that this type will eventually include a lot of data that it does not currently, including the search phrase. (As you noted, the field name will probably also change to reflect this.) For now, sticking with a basic implementation feels to me like a good approach, given that we'll be extending it in the hopefully near future.

PR #13, opened 2023-11-08
Engx242 identifierpatterns

JPrevost:
### Why these changes are being introduced

Adds detection and return of detected ISSN, ISBN, DOI, PMID if requested via GraphQL

If the field is not requested via GraphQL, we do not do the detection.

Not included at this time is external lookups to return additional information, but we may want to discuss how that would work as part of this to ensure that need fits in cleanly with this GraphQL design

### Side effects of this change

* Adds support for lookupTerm to be able to query without logging an event
* Adds `phrase` to SearchEventType even though it is technically part of the TermType for convenience

### Relevant ticket(s)

* https://mitlibraries.atlassian.net/browse/ENGX-242
* https://mitlibraries.atlassian.net/browse/ENGX-241

### DIFF
--- a/app/models/standard_identifiers.rb
+++ b/app/models/standard_identifiers.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+# StandardIdentifiers is a PatternDectector implementation that detects the identifiers DOI, ISBN, ISSN, PMID.
+# See /docs/reference/pattern_detection_and_enhancement.md for details.
+class StandardIdentifiers
+  attr_reader :identifiers
+
+  def initialize(term)
+    @identifiers = {}
+    term_pattern_checker(term)
+    strip_invalid_issns
+  end
+
+  private
+
+  def term_pattern_checker(term)
+    term_patterns.each_pair do |type, pattern|
+      @identifiers[type.to_sym] = match(pattern, term) if match(pattern, term).present?
+    end
+  end
+
+  # Note on the limitations of this implementation
+  # We only detect the first match of each pattern, so a search of "1234-5678 5678-1234" will not return two ISSNs as
+  # might be expected, but just "1234-5678". Using ruby's string.scan(pattern) may be worthwhile if we want to detect
+  # all possible matches instead of just the first. That may require a larger refactor though as initial tests of doing
+  # that change did result in unintended results so it was backed out for now.
+  def match(pattern, term)
+    pattern.match(term).to_s.strip
+  end
+
+  # term_patterns are regex patterns to be applied to the basic search box input
+  def term_patterns
+    {
+      isbn: /\b(ISBN-*(1[03])* *(: ){0,1})*(([0-9Xx][- ]*){13}|([0-9Xx][- ]*){10})\b/,
+      issn: /\b[0-9]{4}-[0-9]{3}[0-9xX]\b/,
+      pmid: /\b((pmid|PMID): (\d{7,8}))\b/,
+      doi: %r{\b10\.(\d+\.*)+/(([^\s.])+\.*)+\b}
+    }
+  end
+
+  def strip_invalid_issns
+    return unless @identifiers[:issn]
+
+    @identifiers[:issn] = nil unless validate_issn(@identifiers[:issn])
+  end
+
+  # validate_issn is only called when the regex for an ISSN has indicated an ISSN
+  # of sufficient format is present - but the regex does not attempt to
+  # validate that the check digit in the ISSN spec is correct. This method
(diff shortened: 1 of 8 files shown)


docs/reference/patterns.md
```diff
@@ -0,0 +1,58 @@
+## Pattern detection and metadata enhancement
+
+The PatternDetector is responsible for identifying specific patterns within the input, such as using regular expressions to detect ISSN, ISBN, DOI, and PMID. Other techniques than regular expressions may also occur here, such as doing phrase matching to identify known scientific journals, or fingerprint matching to identify librarian curated responses.
```
jazairi, line 3: I think `Pattern Detector` is a more descriptive name for what the model and GraphQL type are trying to do, but less descriptive for the GraphQL field than `StandardIdentifiers`. Would it be confusing for the field to have a different name than the model and type? Is that even possible?

This documentation also mentions an `Enhancer`, which is not yet implemented in this PR. (I assume it will be soon, though, based on the experiment you shared last week.) If that model/type is not named `Enhancer`, then an alternative approach could be to clarify which models/types are an example of `Pattern Detector`, and which are an example of `Enhancer`. In other words, `Pattern Detector` and `Enhancer` are how we're describing the features at a high level, whereas `StandardIdentifiers` and `RecordsFound` (or whatever) are the actual implementations of those features. In that case, we should make that clarification in this reference doc.

I do think we should make a decision on this before merging, because I suspect it will be confusing if the naming conventions in the docs don't match what's in the code.

app/models/standard_identifiers.rb
```diff
@@ -0,0 +1,69 @@
...
+  # We only detect the first match of each pattern, so a search of "1234-5678 5678-1234" will not return two ISSNs as
+  # might be expected, but just "1234-5678". Using ruby's string.scan(pattern) may be worthwhile if we want to detect
+  # all possible matches instead of just the first. That may require a larger refactor though as initial tests of doing
+  # that change did result in unintended results so it was backed out for now.
+  def match(pattern, term)
+    pattern.match(term).to_s.strip
+  end
+
+  # term_patterns are regex patterns to be applied to the basic search box input
+  def term_patterns
```
jazairi, line 30: Would it be useful to detect handles? (To be clear, I'm not requesting a change; just curious.)

JPrevost: Interesting idea. I don't know enough about handles to know if this would be useful useful, or just theoretically useful. Before resolving this conversation we can open a ticket to explore this further.

docs/reference/patterns.md
JPrevost: This is super helpful feedback and I agree. I'll give it some thought, likely ask for a synchronous chat, then make appropriate changes.

app/models/standard_identifiers.rb
jazairi: Good idea. I haven't really done any analysis of our search logs, so I think a spike is in order: https://mitlibraries.atlassian.net/browse/ENGX-245

docs/reference/patterns.md
JPrevost: I'm leaning towards having PatternDetectors and Enhancers be the high level concept for this type of work, and StandardIdentifiers being our first implementation of a PatternDetector and coming soon, LookupDoi (etc) being implementations of Enhancers.

I'll be adding a commit that reflects this and then reaching out for further discussion.

docs/reference/pattern_detection_and_enhancement.md
```diff
@@ -1,10 +1,10 @@
 ## Pattern detection and metadata enhancement
 
-The PatternDetector is responsible for identifying specific patterns within the input, such as using regular expressions to detect ISSN, ISBN, DOI, and PMID. Other techniques than regular expressions may also occur here, such as doing phrase matching to identify known scientific journals, or fingerprint matching to identify librarian curated responses.
+A PatternDetector is responsible for identifying specific patterns within the input, such as using regular expressions to detect ISSN, ISBN, DOI, and PMID (implemented in our StandardIdentifiers Class). Other techniques than regular expressions may also occur as PatternDetectors, such as doing phrase matching to identify known scientific journals, or fingerprint matching to identify librarian curated responses.
```
jazairi, line 3: This is non-blocking: it may be useful to write this as `Pattern Detector` rather than `PatternDetector`, so it's clear that we're talking about a concept and not a model/module.

jazairi (approved):
This documentation reads much clearer to me, and I think this approach gives us the best of both naming conventions. Thanks for incorporating that feedback. 👍

PR #18, opened 2023-11-21
Cleanup work

JPrevost:
This PR covers a few small changes grouped into individual commits.

It is likely best to look at each individual commit in this case as they are unrelated, but grouped for convenience to me rather than making 4 PRs. Sorry :) 

The end result should be a functional application running in Heroku (to this point, GraphQL did not work nor did GraphiQL load and the app had no splash page to load).


jazairi (approved):
Thanks for this. Nice to have a real landing page with our theme. :shipit:

PR #40, opened 2024-05-17
Add rack-cors and configuration

matt-bernhardt:
** Why are these changes being introduced:

* The application needs to accept communications from other domains in order to do its job.

** Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/engx-266

** How does this address that need:

* This adds the rack-cors gem, with the recommended default config file, to enable the application to receive communication from our various search applications.

** Document any side effects to this change:

* None

### Why these changes are being introduced

A sentence or two explaining why these changes are here.

### How this addresses that need

Describe the details of the changes.

### Side effects of this change

Are there any side effects? Dependency changes?

### Relevant ticket(s)

* https://mitlibraries.atlassian.net/browse/[X]


README.md
```diff
@@ -6,6 +6,8 @@
 
 `LINKRESOLVER_BASEURL`: base url for our link resolver. `https://mit.primo.exlibrisgroup.com/discovery/openurl?institution=01MIT_INST&rfr_id=info:sid/mit.tacos.api&vid=01MIT_INST:MIT` is probably the best value unless you are doing something interesting.
 
+`ORIGINS`: comma-separated list of domains allowed to connect to (and thus query or contribute to) the application. If not defined, no external connections will be permitted.
```
JPrevost, line 9: Maybe add a note that non-standard ports also need to be included (i.e. `localhost` will not work when running on `localhost:5000`, etc)

JPrevost (approved):
Looks good. I added a note to consider updating the docs to reference a possible local development confusion around port numbers that may not be immediately obvious for developers unfamiliar with CORS.

matt-bernhardt:
Good call - I had to work out the port number thing myself, so that's probably good to have in the readme. I've pushed a commit to update the description there.

JPrevost (approved):

PR #48, opened 2024-06-25
Add ADR to select an authorization gem

jazairi:

We will need authorization in TACOS for the staff UI, which first requires us to choose an authorization gem.

### Relevant ticket(s):

* [TCO-30](https://mitlibraries.atlassian.net/browse/TCO-30)

This adds an ADR to select an authorization gem from among CanCanCan, Pundit, and Action Policy.

The decision we make will affect how we implement authorization.

[TCO-30]: https://mitlibraries.atlassian.net/browse/TCO-30?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/architecture-decisions/0005-use-cancancan-for-authorization.md
+++ b/docs/architecture-decisions/0005-use-cancancan-for-authorization.md
@@ -53,7 +53,11 @@ for a more detailed example.)
 
 Pundit also allows for more granular access control via
 [scopes](https://github.com/varvet/pundit?tab=readme-ov-file#scopes). This feature is the most compelling reason to
-choose Pundit over CanCanCan, but it's unclear whether we would need it in this application.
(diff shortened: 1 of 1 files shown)


JPrevost (approved):
I support the decision to use CanCanCan. Our needs are likely to be simple for the foreseeable future and the additional features/speed of the alternatives do not outweigh the additional complexity/repetitiveness/boilerplate they require. If our needs change over time, I agree with the assessment here that we could migrate to another option when we better understand those needs.

matt-bernhardt (approved):
Thanks for this writeup, and the links you've shared to the alternative gems under consideration.

I'm comfortable with this choice, and definitely find value in having a shared approach to shared problems like authorization across our applications.

:shipit:

PR #55, opened 2024-07-09
Add an ADR for how we approach the administrative interface

matt-bernhardt:
## Why these changes are being introduced:

We need to decide how the application will handle content administration for its internal records. This includes both the records that will be created and managed within the app (fingerprinted phrases for hints, categorization notes), as well as - potentially - records received and stored for use (search terms, imported journal titles, etc).

## Relevant ticket(s):

* [TCO-37](https://mitlibraries.atlassian.net/browse/TCO-37)

## How this addresses that need:

This adds an architecture decision record to survey the available gems for content administration, and proposes a choice for how to proceed (to use Administrate for our initial use cases, but a bias towards leaving it behind rather than heavily customizing it as our needs become clearer).

## Side effects of this change:

This decision might be a bit unorthodox, as it includes the choice to change implementations. That choice may be altered as we build experience with the application, or as the Administrate gem continues to develop.

[TCO-37]: https://mitlibraries.atlassian.net/browse/TCO-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/architecture-decisions/0008-choose-administrative-ui-strategy.md
+++ b/docs/architecture-decisions/0008-choose-administrative-ui-strategy.md
@@ -1,0 +1,76 @@
+# 8. Choose administrative UI strategy
+
+Date: 2024-07-11
+
+## Status
+
+Accepted
+
+## Context
+
+Part of the TACOS application involves the maintenance of records: specific
+phrases that should trigger bespoke responses, notes received from staff who
+participate in the categorization process, etc. Other records, like the set of
+search terms, any derived metadata from those terms, and imported records like
+journal titles, might be viewable at least (if not editable).
+
+Because of these needs, the application has a need for an administrative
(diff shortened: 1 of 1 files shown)


JPrevost (approved):
Agreed.

Using Administrate for anything it works for out-of-the-box and building our own interfaces whenever it isn't what we need is a good plan... as long as we hold ourselves to it in the future :)

jazairi (approved):
I agree with the choice of Administrate here. It has its issues, but I'm sure that Active Admin, Rails Admin, Avo, et al all do as well. And we're familiar with Administrate, warts and all.

I also appreciate the warning against template customization. My sense is that, as soon as we start customizing _any_ template, we should instead build that dashboard from scratch.

PR #57, opened 2024-07-11
Add ADR to select an authentication approach

jazairi:

We will need authentication in TACOS for the staff UI.

#### Relevant ticket(s):

* [TCO-28](https://mitlibraries.atlassian.net/browse/TCO-28)

This adds an ADR to select an authentication approach from among Devise, Clearance, Sorcery, AuthLogic, and a custom solution using built-in Rails security features.

The decision we make will affect how we implement authentication.

[TCO-28]: https://mitlibraries.atlassian.net/browse/TCO-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/architecture-decisions/0007-use-devise-for-authentication.md
+++ b/docs/architecture-decisions/0007-use-devise-for-authentication.md
@@ -1,0 +1,104 @@
+# 7. Use Devise for authentication
+
+Date: 2024-07-11
+
+## Status
+
+Accepted
+
+## Context
+
+We will need to authenticate users in TACOS to help manage access to the staff interface. The ideal authentication
+approach would involve Omniauth integration, as IS&T recently implemented Okta, and
+[a strategy exists](https://github.com/omniauth/omniauth_openid_connect) to authenticate with OIDC using Omniauth.
+
+## Options considered
+
+### [Devise](https://github.com/heartcombo/devise)
+
+Devise is the far and away the most popular authentication gem, and the one with which we have the greatest familiarity.
+It is well-documented and feature-rich, and it integrates seamlessly with Omniauth, which we will likely use to work
+with Okta.
+
+Devise is also something of a black box, and its extensibility causes a steep learning curve. All of us in EngX have
+worked with Devise, but not all of us have configured it from scratch in a new application. As a result, our collective
+understanding of how the gem works is fairly limited.
+
+We also anticipate fairly straightforward authentication needs for this application, so its extensibility may just add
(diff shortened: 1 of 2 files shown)


docs/architecture-decisions/0007-use-devise-for-authentication.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+We will need to authenticate users in TACOS to help manage access to the staff interface. The ideal authentication
+approach would involve Omniauth integration, as IS&T recently implemented Okta, and
+[a strategy already exists](https://github.com/omniauth/omniauth-okta) to authenticate in Okta using Omniauth.
```
JPrevost, line 13: I don't think we should use that omniauth integration as it seems unsupported, https://github.com/omniauth/omniauth_openid_connect is more likely the path forward.

docs/architecture-decisions/0007-use-devise-for-authentication.md
```diff
@@ -0,0 +1,104 @@
...
+implement due to their lack of Omniauth integration.
+
+Authlogic feels equally complex as Devise, albeit more transparent. It also seems not to be especially well maintained.
+
+A custom solution would be an interesting way leverage built-in Rails functionality while reducing our reliance on
+third-party libraries. However, it is difficult to justify considering the additional risk and development time.
+
+Ultimately, while all of these alternatives are appealing, there is a distinct advantage to choosing a gem that "just
+works" and is widely adopted. Much like our
+[decision to use CanCanCan for authorization](https://github.com/MITLibraries/tacos/blob/main/docs/architecture-decisions/0005-use-cancancan-for-authorization.md), I might not feel like Devise is the best authentication gem, but I have
```
JPrevost, line 96: I think this will need an updated URL as that linked ADR is changing to 0006

JPrevost (approved):
Solid write up, thanks!

I support the decision in this ADR as written 👍🏻

matt-bernhardt (approved):
Thanks for this write-up. While I'm not separately familiar with the options for authentication gems, I agree with the framing you've provided, and support the decision you reach based on that framing.

While in general I like to build things ourselves, there are certain domains where it feels better to trust in the experience of others. Authentication is one of those domains.

:shipit:

PR #61, opened 2024-07-19
Tco48 journal data loader

JPrevost:

* Loading journals from an external source is how we'll be internally
  detecting journal name matches in this experimental detector.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-41
* https://mitlibraries.atlassian.net/browse/TCO-48

* Adds data harvester from OpenAlex Sources API endpoint
* Adds data loader from the result of the OpenAlex harvest which can
  load from local or remote file

* Technically there are more than Journals being loaded in the current
query configuration from OpenAlex. We are pulling both Journals and
Book Series. It felt like Journals was still a fine internal name for
this but we could also consider Serials to be more accurate.

* Remote file loading only handles URI hosted files. If we move the
OpenAlex Harvester functionality into a data pipeline that exports into
S3 directly in the future we may want to add S3 support to the loader
rather than exposing the S3 files via https (although the data is open
so there is no harm in exposing direclty via https so it we could choose
either)

* Removes filtering of SSN as it also catches ISSN (see separate commit message)


config/initializers/filter_parameter_logging.rb
```diff
@@ -3,6 +3,6 @@
 # Configure parameters to be partially matched (e.g. passw matches password) and filtered from the log file.
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
-Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+Rails.application.config.filter_parameters += %i[
+  passw secret token _key crypt salt certificate otp
 ]
```
JPrevost, line 8: Most of the change was auto-formatted when I resaved to match our current best practices. The only practical change was removal of `:ssn` from the list of filters.

lib/tasks/journals.rake
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+# Loaders can bulk load data
+namespace :journals do
+  # openalex_harvester uses the OpenAleX Sources API endpoint to harvest to a local json file
+  #
+  # @note currently, we are limititing to just core Sources but not limiting to journals. We may want to consider
```
jazairi, line 7: Typo: limiting

lib/tasks/journals.rake
```diff
@@ -0,0 +1,110 @@
...
+    if URI(args.file).scheme
+      Rails.logger.info("Loading data from remote file #{args.file}")
+      data = URI.open(args.file, 'rb', &:read)
+    else
+      Rails.logger.info("Loading data from local file #{args.file}")
+      data = File.read(args.file)
+    end
+
+    # Delete all journals. We do this to simplify the loader process to avoid consideration of updates/deletes.
+    Detector::Journal.delete_all
```
jazairi, line 96: Is there an edge case in which we _wouldn't_ want to delete everything first? I'm guessing not, but this does worry me a bit.

jazairi (approved):
I confirmed that the rake tasks work as expected. Struggling with a headache this afternoon, so this hasn't been the most thorough review. (Mostly, I was hoping to better understand what cursors do before approving, but it works and I trust your judgment.)

Both comments are non-blocking. The `Detector::Journal.delete_all` call gives me some pause, but I couldn't think of a reason why we couldn't clear that table each time.

JPrevost:
@jazairi cursors are cool. TIMDEX definitely needs them and OpenSearch already supports them so we can add them easily(ish). Basically it's just more efficient pagination and addresses the problem of how to page past 10,000ish records in a search without needing to calculate each preceding page before showing the requested page. They are temporary in nature and allow for getting "all records that matched your query" in a neat way.

PR #65, opened 2024-07-24
Adds exact journal matches to historical analysis

JPrevost:

* Understanding our ability to detect search intent over time is core to our ability to know how we are doing

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-54

* Adds new field to track exact journal matches to Metrics::Algorithms
* Updates Metrics::Algorithms to run journal exact matches in addition to the existing StandardIdentifer matches. This included a refactor to better support multiple match types.

Journal exact matching is not guaranteed to be an indicator of user search intent because Journal names are also common words in many cases. When we build our our validation workflows, we'll be able to understand what percentage of these types of matches are definitely Journals and what percentage is ambiguous. We can likely update our algorithm to drop some of the more ambiguous detections at that point.

### DIFF
--- a/test/models/metrics/algorithms_test.rb
+++ b/test/models/metrics/algorithms_test.rb
@@ -4,15 +4,16 @@
 #
 # Table name: metrics_algorithms
 #
-#  id         :integer          not null, primary key
-#  month      :date
-#  doi        :integer
-#  issn       :integer
-#  isbn       :integer
-#  pmid       :integer
-#  unmatched  :integer
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
+#  id            :integer          not null, primary key
+#  month         :date
+#  doi           :integer
+#  issn          :integer
+#  isbn          :integer
+#  pmid          :integer
+#  unmatched     :integer
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  journal_exact :integer
 #
 require 'test_helper'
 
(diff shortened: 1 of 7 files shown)


test/models/metrics/algorithms_test.rb
```diff
@@ -38,6 +39,11 @@ class Algorithms < ActiveSupport::TestCase
     assert aggregate.pmid == 1
   end
 
+  test 'journal exact counts are included in monthly aggregation' do
+    aggregate = Metrics::Algorithms.new.generate(DateTime.now)
+    assert aggregate.journal_exact == 1
```
jazairi, line 44: I worry a bit that this type of testing will lead to similar issues as we've seen in ETD (i.e.,adding a fixture breaks a bunch of tests). Not requesting a change, partly because I don't have any good suggestions and partly because I'd love to get this ticket to done.

jazairi (approved):
Exciting feature, thanks for this! While I agree that the current iteration of exact title matching may not be predictive of user intent, I think this data will help us better understand what our users are searching for (and how to refine the algorithm).

test/models/metrics/algorithms_test.rb
JPrevost: Yeah, you are definitely correct. As more fixtures are added, this test could fail.

I don't have a great idea that doesn't involve just not using fixtures or reimplementing the algorithm that calculates matches in the test itself... neither of which are really acceptable.

Thanks for pointing this out as I do think it's important for us to acknowledge when these "future" problems are being introduced.

PR #66, opened 2024-07-24
Add Detector::SuggestedResource model

matt-bernhardt:
## Developer

This adds a basic SuggestedResource model, under the Detector namespace, along with logic for calculating and maintaining the fingerprint for these records.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-22

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

YES dependencies are updated
YES migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/detector/suggested_resource.rb
+++ b/app/models/detector/suggested_resource.rb
@@ -10,5 +10,41 @@
 #  created_at  :datetime         not null
 #  updated_at  :datetime         not null
 #
+
+require 'stringex/core_ext'
+
 class Detector::SuggestedResource < ApplicationRecord
+  before_save :update_fingerprint
+
+  def update_fingerprint
+    self.fingerprint = calculate_fingerprint
+  end
+
+  # This implements the OpenRefine fingerprinting algorithm. See
+  # https://openrefine.org/docs/technical-reference/clustering-in-depth#fingerprint
+  def calculate_fingerprint
+    temp = self.phrase
+    temp.strip!
+    temp.downcase!
+
+    # This removes all punctuation and symbol characters from the string.
+    temp.gsub!(/\p{P}|\p{S}/, '')
+
+    # Normalize to ASCII (e.g. gödel and godel are liable to be intended to
+    # find the same thing)
+    temp = temp.to_ascii
+
+    # Coercion to ASCII can introduce new symbols, so we remove those now.
+    temp.gsub!(/\p{P}|\p{S}/, '')
+
+    # Tokenize
+    array = temp.split
+
+    # Remove duplicates and sort
+    array.uniq!
+    array.sort!
+
+    # Rejoin tokens
+    new_fingerprint = array.join(' ')
+  end
 end
--- a/app/models/detector/suggested_resource.rb
+++ b/app/models/detector/suggested_resource.rb
@@ -1,0 +1,2 @@
+class Detector::SuggestedResource < ApplicationRecord
+end
--- a/app/models/detector/suggested_resource.rb
+++ b/app/models/detector/suggested_resource.rb
@@ -23,35 +23,35 @@ module Detector
     before_save :update_fingerprint
 
     def update_fingerprint
-      self.fingerprint = calculate_fingerprint
+      self.fingerprint = calculate_fingerprint(phrase)
     end
 
     # This implements the OpenRefine fingerprinting algorithm. See
     # https://openrefine.org/docs/technical-reference/clustering-in-depth#fingerprint
-    def calculate_fingerprint
-      temp = phrase
-      temp.strip!
-      temp.downcase!
+    def calculate_fingerprint(old_phrase)
+      modified_phrase = old_phrase
+      modified_phrase = modified_phrase.strip
+      modified_phrase = modified_phrase.downcase
 
       # This removes all punctuation and symbol characters from the string.
-      temp.gsub!(/\p{P}|\p{S}/, '')
+      modified_phrase = modified_phrase.gsub(/\p{P}|\p{S}/, '')
 
       # Normalize to ASCII (e.g. gödel and godel are liable to be intended to
       # find the same thing)
-      temp = temp.to_ascii
+      modified_phrase = modified_phrase.to_ascii
 
       # Coercion to ASCII can introduce new symbols, so we remove those now.
-      temp.gsub!(/\p{P}|\p{S}/, '')
+      modified_phrase = modified_phrase.gsub(/\p{P}|\p{S}/, '')
 
       # Tokenize
-      array = temp.split
+      tokens = modified_phrase.split
 
       # Remove duplicates and sort
-      array.uniq!
-      array.sort!
+      tokens = tokens.uniq
+      tokens = tokens.sort
 
       # Rejoin tokens
-      array.join(' ')
+      tokens.join(' ')
     end
   end
 end
--- a/app/models/detector/suggested_resource.rb
+++ b/app/models/detector/suggested_resource.rb
@@ -1,2 +1,14 @@
+# == Schema Information
+#
+# Table name: detector_suggested_resources
+#
+#  id          :integer          not null, primary key
+#  title       :string
+#  url         :string
+#  phrase      :string
+#  fingerprint :string
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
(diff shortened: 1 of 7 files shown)


app/models/detector/suggested_resource.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      # Normalize to ASCII (e.g. gödel and godel are liable to be intended to
+      # find the same thing)
+      temp = temp.to_ascii
+
+      # Coercion to ASCII can introduce new symbols, so we remove those now.
+      temp.gsub!(/\p{P}|\p{S}/, '')
+
+      # Tokenize
+      array = temp.split
```
JPrevost, line 47: Non-blocking. `array` is not my favorite name for this. What's an array of? The comment mentions it is creating tokens. Maybe `tokens` would be a better name rather than naming it after it's type?

app/models/detector/suggested_resource.rb
```diff
@@ -0,0 +1,57 @@
...
+  class SuggestedResource < ApplicationRecord
+    before_save :update_fingerprint
+
+    def update_fingerprint
+      self.fingerprint = calculate_fingerprint
+    end
+
+    # This implements the OpenRefine fingerprinting algorithm. See
+    # https://openrefine.org/docs/technical-reference/clustering-in-depth#fingerprint
+    def calculate_fingerprint
```
JPrevost, line 31: Slightly forward thinking and thus non-blocking, but having the `phrase` be an argument would allow this to be used on incoming search terms rather than only on instantiated `SuggestedResource`s.

app/models/detector/suggested_resource.rb
```diff
@@ -0,0 +1,57 @@
...
+      temp = phrase
```
JPrevost, line 32: Non-blocking: `temp` is not a super useful name for a variable. As we are working with modifying phrases, maybe something like `modifified_phrase` would help us out.

You also use a bunch of `!` operators which change the underlying string rather than creating a copy so this might not be required at all (I believe phrase.strip! is the same as temp = phrase; temp.strip!, etc)

app/models/detector/suggested_resource.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      # Tokenize
+      array = temp.split
+
+      # Remove duplicates and sort
+      array.uniq!
+      array.sort!
+
+      # Rejoin tokens
+      array.join(' ')
```
JPrevost, line 54: Non-blocking: I appreciate that it's very clear to see the entire process in this one method, but I'm curious if we'd get better naming by abstracting some of this logic into separate methods. It's very minor and I'm not sure I'd actually do it, but this method ends up doing quite a lot of very small things. I wonder if 2 or 3 methods would allow each to be named clearly and fully documented tested in a way that would be worth the extra "noise" the added complexity would introduce.

db/migrate/20240716143850_create_detector_suggested_resources.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :detector_suggested_resources do |t|
+      t.string :title
+      t.string :url
+      t.string :phrase
+      t.string :fingerprint
+
+      t.timestamps
+    end
+    add_index :detector_suggested_resources, :phrase, unique: true
+    add_index :detector_suggested_resources, :fingerprint, unique: true
```
JPrevost, line 12: Woohoo, indexes!

test/models/detector/suggested_resource_test.rb
```diff
@@ -0,0 +1,68 @@
...
+      refute resource.fingerprint == 'A brand new phrase'
+
+      resource.phrase = 'This is a brand new phrase'
+      resource.save
+      resource.reload
+
+      assert resource.fingerprint == 'a brand is new phrase this'
+    end
+
+    test 'fingerprints standardize characters used' do
```
JPrevost, line 42: This test is covering a lot of functionality. It might be better to split it up into tests for each aspect of the fingerprint algorithm. I think some functionality may not have been included in the testing because it all got put in one. Notably missing (I think?) is de-duplication of the tokens. This test could be updated to include that, or we could have one test per "thing" the algorithm does.

JPrevost (commented):
I'm neither approving or blocking this request. I had a fair amount of non-blocking comments so I'm curious if you'd like to address any of them. If you don't, I'd likely approve but as there may be a bunch of changes (or zero!) I'd like to hold off until I understand how/if you'd like to change anything.

test/models/detector/suggested_resource_test.rb
matt-bernhardt: Yeah, I was tempted to split the tests into different features. I'm open to going that route, to be more explicit about what the different components are.

app/models/detector/suggested_resource.rb
matt-bernhardt: While I appreciate the value of shorter methods, and this one is currently a bit long, I'm not sure what the natural breaking points would be. There's a "too far" region where we're defining our own methods to match those of the tools we're using, which feels like it would force us to refactor heavily if we ever change the workflow at all.

That said, one natural split point would seem to be around the tokenization step, maybe? One method to prep the submitted phrase, and then one method to combine and reorder everything?

matt-bernhardt: Another consideration - is there ever going to be a point where we'd need to call only one of these potential methods? I struggle to think of one.

app/models/detector/suggested_resource.rb
matt-bernhardt: In general, I'm not going to defend the variable names in this method. I'm not sure at all about just applying these steps to `phrase`, though - it feels like that would alter the information we got from staff, and which they'll be looking for when they try to maintain these records.

JPrevost: Oh, I don't particularly love the bang methods at all here. I'm just saying if you are going to use them, go nuts with them because you have determined it is safe to do so. They aren't necessary at all if you are assigning the output to a variable. Their entire point is to change the thing _in place_ unsafely.

From the docs:
```
The bang methods (! at the end of the method name) are called and executed just like any other method. However, by convention, a method with an exclamation point or bang is considered dangerous. In Ruby's core library the dangerous method implies that when a method ends with a bang (!), it indicates that unlike its non-bang equivalent, permanently modifies its receiver. Almost always, the Ruby core library will have a non-bang counterpart (method name which does NOT end with !) of every bang method (method name which does end with !) that does not modify the receiver. This convention is typically true for the Ruby core library but may or may not hold true for other Ruby libraries.
```

So it is entirely possible you _are_ modifying `phrase` but just didn't know it but using the bangs.

```ruby
irb(main):038:0> x = "HALLO"
=> "HALLO"
irb(main):039:0> y = x.downcase!
=> "hallo"
irb(main):040:0> x
=> "hallo"

```

JPrevost: Essentially, if you want to safely use bangs here you should pass a string in as an argument rather than using the `phrase` attribute of the model.

app/models/detector/suggested_resource.rb
JPrevost: Regardless of whether we ever call these pieces independently, it is easier to _read_ the method if we create well named small methods to split up complex processes. I do agree that this is right on the edge of being complex enough to make a difference. If you use better naming within this method as suggested in other comments, I suspect my inclination to split it into smaller methods would decline.

matt-bernhardt:
Okay, @JPrevost - I think this is now ready for review.

test/models/detector/suggested_resource_test.rb
```diff
@@ -0,0 +1,123 @@
...
+    test 'fingerprints are generated automatically' do
+      resource = {
+        title: 'Our latest resource',
+        url: 'https://example.org',
+        phrase: 'Our latest resource'
+      }
+
+      new_resource = Detector::SuggestedResource.create(resource)
+
+      assert new_resource.fingerprint == 'latest our resource'
```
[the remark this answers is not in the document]
matt-bernhardt: Ahh... thanks for this. I'm sorely tempted to make this change now, but I think I'd rather just do the refactoring as part of adopting the `rubocop-minitest` gem. We have that in place for some of our other apps, right? ISTR getting feedback about assertions in the past.

PR #67, opened 2024-07-26
Implement authentication

jazairi:
This adds Touchstone authentication via Devise and Omniauth, along with fake auth for local development and PR builds. The initial auth setup and the fake auth config are in separate commits for ease of review. Please check each commit message for more detailed info on the changesets.

For the time being, I've opted to use Omniauth developer mode for fake auth. We might benefit from using Keycloak in local dev to better simulate a real authentication flow, but setting that proved somewhat complicated and is probably best saved for another time. I've created [a spike ticket](https://mitlibraries.atlassian.net/browse/TCO-59) to investigate further.

To confirm functionality:
* Fake auth logic can be tested in this PR app and locally. It's best to confirm both, because the logic that triggers fake auth is different for each environment. (Local checks for the development environment, whereas PR builds check for the `HEROKU_APP_NAME` env.)
* Prod auth can be tested in staging. I've been doing that by deploying this branch, confirming that I can log in, then redeploying `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,24 +16,36 @@ To see a current list of commands, run `make help`.
 
 `UNPAYWALL_EMAIL`: email address to include in API call as required in their [documentation](https://unpaywall.org/products/api). Your personal email is appropriate for development. Deployed and for tests, use the timdex moira list email.
 
-#### Authentication
+### Optional
+
+`PLATFORM_NAME`: The value set is added to the header after the MIT Libraries logo. The logic and CSS for this comes
+from our theme gem.
+
+### Authentication
+
+#### Required in all environments
 
 Access to some of the config values below is limited. Please contact someone in the EngX team if you need help locating
 them.
 
 `BASE_URL`: The base url for the app. This is required for Omniauth config.
-`FAKE_AUTH_CONFIG`: Switches Omniauth to developer mode when set to `true`. Fake auth is also enabled whenever the
-app is started in the development environment, or if the app is a PR build (see `HEROKU_APP_NAME` under optional
-variables).
-`OP_HOST`: The OID provider hostname, required for authentication. (Do not include URL prefix.)
-`OP_SECRET_KEY`: The secret key for the OID client.
-`OP_CLIENT_ID`: The identifier for the OID client.
-`OP_ISSUER`: The URL for the OIDC issuer. This can be found in the Touchstone OpenID metadata.
+`OPENID_HOST`: The OID provider hostname, required for authentication. (Do not include URL prefix.)
+`OPENID_SECRET_KEY`: The secret key for the OID client.
+`OPENID_CLIENT_ID`: The identifier for the OID client.
+`OPENID_ISSUER`: The URL for the OIDC issuer. This can be found in the Touchstone OpenID metadata.
 
-### Optional
+#### Required in PR builds
 
-`HEROKU_APP_NAME`: Used by the FakeAuthConfig module to determine whether an app is a PR build.
-`PLATFORM_NAME`: The value set is added to the header after the MIT Libraries logo. The logic and CSS for this comes from our theme gem.
+The config below is needed to run Omniauth in developer mode in Heroku review apps. Rather than relying upon a single
+ENV value, we use the `FakeAuthConfig` module to perform additional checks that confirm whether developer mode should
+be enabled. This assures that developer mode is never enabled in staging or production apps.
+
+`FAKE_AUTH_ENABLED`: Switches Omniauth to developer mode when set. If unset, PR builds will attempt to authenticate with
+OIDC, which will fail as their domains are not registered with the provider. (Note: Developer mode is also enabled
+whenever the app is started in the development environment.)
+`HEROKU_APP_NAME`: Used by the FakeAuthConfig module to determine whether an app is a PR build. If this is set along
+with `FAKE_AUTH_ENABLED`, then Omniauth will use Developer mode. Heroku sets this variable automatically for review
+apps; it should never be manually set or overridden in any environment.
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ Access to some of the config values below is limited. Please contact someone in 
 them.
 
 `BASE_URL`: The base url for the app. This is required for Omniauth config.
+`FAKE_AUTH_CONFIG`: Switches Omniauth to developer mode when set to `true`. Fake auth is also enabled whenever the
+app is started in the development environment, or if the app is a PR build (see `HEROKU_APP_NAME` under optional
+variables).
 `OP_HOST`: The OID provider hostname, required for authentication. (Do not include URL prefix.)
 `OP_SECRET_KEY`: The secret key for the OID client.
 `OP_CLIENT_ID`: The identifier for the OID client.
@@ -29,6 +32,7 @@ them.
 
 ### Optional
 
+`HEROKU_APP_NAME`: Used by the FakeAuthConfig module to determine whether an app is a PR build.
 `PLATFORM_NAME`: The value set is added to the header after the MIT Libraries logo. The logic and CSS for this comes from our theme gem.
 
 ## Documentation
--- a/README.md
+++ b/README.md
@@ -16,6 +16,17 @@ To see a current list of commands, run `make help`.
 
 `UNPAYWALL_EMAIL`: email address to include in API call as required in their [documentation](https://unpaywall.org/products/api). Your personal email is appropriate for development. Deployed and for tests, use the timdex moira list email.
 
+#### Authentication
+
+Access to some of the config values below is limited. Please contact someone in the EngX team if you need help locating
+them.
+
+`BASE_URL`: The base url for the app. This is required for Omniauth config.
+`OP_HOST`: The OID provider hostname, required for authentication. (Do not include URL prefix.)
+`OP_SECRET_KEY`: The secret key for the OID client.
+`OP_CLIENT_ID`: The identifier for the OID client.
+`OP_ISSUER`: The URL for the OIDC issuer. This can be found in the Touchstone OpenID metadata.
+
 ### Optional
 
 `PLATFORM_NAME`: The value set is added to the header after the MIT Libraries logo. The logic and CSS for this comes from our theme gem.
--- a/app/models/concerns/fake_auth_config.rb
+++ b/app/models/concerns/fake_auth_config.rb
@@ -1,0 +1,27 @@
+module FakeAuthConfig
+  # Used in an initializer to determine if the application is configured and allowed to use fake authentication.
+  def self.fake_auth_status
+    return true if fake_auth_enabled? && app_name_pattern_match?
+
+    false
+  end
+
+  # Default to fake auth in development unless FAKE_AUTH_ENABLED=false. This allows rake tasks to run without loading
+  # ENV.
+  private_class_method def self.fake_auth_enabled?
+    if Rails.env.development? && ENV['FAKE_AUTH_ENABLED'].nil?
+      true
+    else
+      ENV['FAKE_AUTH_ENABLED'] == 'true'
+    end
+  end
+
+  # Checks if the current app is a Heroku pipeline app, in which case fake_auth should be enabled.
+  # In test ENV we require setting a fake app name to allow for testing of the pattern.
+  private_class_method def self.app_name_pattern_match?
+    return true if Rails.env.development?
+
+    review_app_pattern = /^tacos-api-pipeline-pr-\d+$/
+    review_app_pattern.match(ENV.fetch('HEROKU_APP_NAME', nil)).present?
+  end
+end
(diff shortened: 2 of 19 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_07_01_205444) do
+ActiveRecord::Schema[7.1].define(version: 2024_07_15_190838) do
```
JPrevost, line 13: @jazairi noting this here even though you aren't in review formally yet because I just ran into a conflict like this on the branch I merged. Your timestamp here is before the current timestamp in `main` and may cause issues and lead to migrations being skipped. 

For me, my timestamp was larger than main, so I was able to use it without worry. Yours is older which may cause issues if you select yours (or main?). It's unclear if the best path forward is to create a new timestamp or go with the higher number already in main.

jazairi: Going with what's in `main` worked locally, but I suppose we'll see what happens when this lands...

jazairi:
At this point, I'm comfortable putting this into review. The test coverage drop is related to the Omniauth developer strategy details, which we also don't test in ETD. To be clear, though, I'd be open to testing any/all of that here if the reviewer feels it's appropriate.

config/locales/devise.en.yml
```diff
@@ -0,0 +1,65 @@
+# Additional translations at https://github.com/heartcombo/devise/wiki/I18n
```
JPrevost, line 1: Was this added as part of the devise generator? 99% of it seems irrelevant is why I ask, but if it was generated I'm fine with it(ish).

config/routes.rb
```diff
@@ -1,4 +1,9 @@
 Rails.application.routes.draw do
+  devise_for :users, controllers: { omniauth_callbacks: 'users/omniauth_callbacks' }, only: [:omniauth_callbacks]
```
JPrevost, line 2: It's interesting that `user_developer_omniauth_authorize_path`  (and the OIDC variant) are not defined in routes.

test/integration/authentication_test.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def silence_omniauth
+    previous_logger = OmniAuth.config.logger
+    OmniAuth.config.logger = Logger.new('/dev/null')
+    yield
+  ensure
+    OmniAuth.config.logger = previous_logger
+  end
+
+  test 'accessing callback without credentials redirects to signin' do
```
JPrevost, line 20: I'm not sure this tests for what it says it tests for. It says it should be redirected to the sign in path but only confirms any 200 was received right?

test/models/fake_auth_test.rb
```diff
@@ -0,0 +1,55 @@
+require 'test_helper'
+
+class FakeAuthTest < ActiveSupport::TestCase
+  include FakeAuthConfig
+
+  test 'fakeauth disabled' do
+    ClimateControl.modify(
+      FAKE_AUTH_ENABLED: 'false',
+      HEROKU_APP_NAME: 'thesis-submit-pr-123'
```
JPrevost, line 9: Probably better to use a tacos pattern to confirm specifically that `FAKE_AUTH_ENABLED` being false is enough to shut it off even if `HEROKU_APP_NAME` would be allowed for these two assertions.

test/models/fake_auth_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    ) do
+      assert_equal(false, FakeAuthConfig.fake_auth_status)
+    end
+    ClimateControl.modify(
+      FAKE_AUTH_ENABLED: 'false',
+      HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
```
JPrevost, line 15: See above

config/initializers/_fake_auth_configuration.rb
```diff
@@ -0,0 +1,5 @@
+# This configuration needs to load before the devise initializer so this value
+# is available when needed.
+
+require "#{Rails.root}/app/models/concerns/fake_auth_config.rb"
+Rails.application.config.fake_auth_enabled = FakeAuthConfig.fake_auth_status
```
JPrevost, line 5: I'm somewhat curious why we used `Rails.application.config.fake_auth_enabled` instead of just calling `FakeAuthConfig.fake_auth_status` throughout. I know this is an ETD pattern carried over, but I'm not currently understanding the added value fwiw.

README.md
```diff
@@ -16,8 +16,23 @@ To see a current list of commands, run `make help`.
 
 `UNPAYWALL_EMAIL`: email address to include in API call as required in their [documentation](https://unpaywall.org/products/api). Your personal email is appropriate for development. Deployed and for tests, use the timdex moira list email.
 
+#### Authentication
+
+Access to some of the config values below is limited. Please contact someone in the EngX team if you need help locating
+them.
+
+`BASE_URL`: The base url for the app. This is required for Omniauth config.
+`FAKE_AUTH_CONFIG`: Switches Omniauth to developer mode when set to `true`. Fake auth is also enabled whenever the
```
JPrevost, line 25: I believe this should be `FAKE_AUTH_ENABLED` and it also reads like it is not required in PR builds, but in my testing I had to set `FAKE_AUTH_ENABLED` to true in the PR build to switch the Sign In button to fake auth

JPrevost (commented):
This is one of those reviews in which there isn't really a ton of code but a heck of a lot is going on. Great job!

- OIDC via state: works as expected
- fake auth in PR build: only worked when I added `FAKE_AUTH_ENABLED=true` which is contrary to the docs (both in what variable is involved and that it doesn't need setting)
- worked as advertised (no config changes, fake auth was working, users created as expected)

One test feels a bit misleading, but nothing else is blocking.

README.md
```diff
@@ -16,8 +16,23 @@ To see a current list of commands, run `make help`.
...
+app is started in the development environment, or if the app is a PR build (see `HEROKU_APP_NAME` under optional
+variables).
+`OP_HOST`: The OID provider hostname, required for authentication. (Do not include URL prefix.)
```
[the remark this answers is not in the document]
jazairi: Totally fair. `OP` is idiomatic (i.e., OID provider), but idiomatic isn't especially helpful if it's not meaningful to us. 🙂

README.md
```diff
@@ -16,8 +16,23 @@ To see a current list of commands, run `make help`.
...
+`OP_SECRET_KEY`: The secret key for the OID client.
+`OP_CLIENT_ID`: The identifier for the OID client.
+`OP_ISSUER`: The URL for the OIDC issuer. This can be found in the Touchstone OpenID metadata.
+
 ### Optional
 
+`HEROKU_APP_NAME`: Used by the FakeAuthConfig module to determine whether an app is a PR build.
```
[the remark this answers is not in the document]
jazairi: Yeah, it's automatically added to PR builds. That's a good call, I'll update this to clarify.

README.md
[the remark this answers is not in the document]
JPrevost: I very likely just spend too much time on Reddit. OP is fine lol

app/models/concerns/fake_auth_config.rb
```diff
@@ -0,0 +1,27 @@
...
+  # Used in an initializer to determine if the application is configured and allowed to use fake authentication.
+  def self.fake_auth_status
+    return true if fake_auth_enabled? && app_name_pattern_match?
+
+    false
+  end
+
+  # Default to fake auth in development unless FAKE_AUTH_ENABLED=false. This allows rake tasks to run without loading
+  # ENV.
+  private_class_method def self.fake_auth_enabled?
```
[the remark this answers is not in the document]
jazairi: I actually changed it to `private`, but rubocop got mad at me. Apparently, `private` doesn't work for class methods. (I think it should, but who am I to say!)

config/locales/devise.en.yml
jazairi: Yup, that's a generated file. Happy to remove or trim it if you think it will be confusing. (We don't use it in ETD.)

test/integration/authentication_test.rb
jazairi: I agree, it could be clearer. We are following a redirect, but not confirming that we land anywhere in particular. I think I'll actually rewrite this to confirm that a new user isn't created.

test/models/fake_auth_test.rb
jazairi: What, you're criticizing my copypasta? 😅

app/models/concerns/fake_auth_config.rb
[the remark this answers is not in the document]
JPrevost: Oh right... makes sense now. I was missing that these were all class methods and just focused on "why is this weird" when it turns out it is not weird, we just normally don't make class methods private because, well, maybe that's the weird bit going on here lol. It's totally fine as we know it works and is never actually run in prod anyway so :shrug:

config/locales/devise.en.yml
JPrevost: Nah, it's fine. Thanks for confirming though.

test/integration/authentication_test.rb
JPrevost: The docs imply we can actually test what the name says it is testing:

https://api.rubyonrails.org/v7.1.3.4/classes/ActionDispatch/IntegrationTest.html (i.e. that we are on a signin path?)

If not, renaming the test to say what we are testing would be great.

config/initializers/_fake_auth_configuration.rb
jazairi: It does feel DRYer to me to do it in an initializer. That way, we don't have to include the module everywhere we use it. That's only in three places currently, so ymmv as to how much value it adds in practice.

One thing I did consider while copying this logic over, and am considering more seriously now, is whether we really need the `FakeAuthConfig` module at all. What's the real value of using that over, say, a Flipflop feature? We basically would just need to add `ENV['FAKE_AUTH_CONFIG']` (or `ENV['FAKE_AUTH_ENABLED']` as it will so be named) in the pipeline and in our local dev environments, right? That doesn't feel any different from how we handle, e.g., API endpoints across multiple environments.

jazairi: Following some additional discussion in Slack, I think I prefer using the initializer and calling `Rails.application.config.fake_auth_enabled` rather than `FakeAuthConfig` for consistency's sake.

test/integration/authentication_test.rb
jazairi: I guess I'm not as concerned about the path as I am that logging in with invalid credentials doesn't create a user. So, I think renaming with some additional logic is the play here, unless you feel otherwise.

JPrevost: Works for me

config/initializers/_fake_auth_configuration.rb
jazairi: Further amending my take on this. 🫠 I'm dropping the initializer due to the filenaming kludge required (i.e., it must load before the devise initializer so the config will be available there). It feels less wonky to me to require the FakeAuthConfig module where we need it.

jazairi:
@JPrevost I think the latest commit addresses all of your feedback, but let me know if I missed something. Ultimately, I decided to remove the fake auth initializer. This branch is currently deployed to staging if you'd like to confirm OIDC still works.

Also, while ENV has been updated in staging, I'm waiting to update in prod until this passes review. (Saying that mostly as a note to myself do it before I merge.)

README.md
```diff
@@ -16,24 +16,35 @@ To see a current list of commands, run `make help`.
...
+#### Required in PR builds
+
+The config below is needed to run Omniauth in developer mode in Heroku review apps. Rather than relying upon a single
+ENV value, we use the `FakeAuthConfig` module to perform additional checks that confirm whether developer mode should
+be enabled. This assures that developer mode is never enabled in staging or production apps.
+
+`FAKE_AUTH_ENABLED`: Switches Omniauth to developer mode when set. If unset, PR builds will attempt to authenticate with
+OIDC, which will fail as their domains are not registered with the provider. (Note: Developer mode is also enabled
+whenever the app is started in the development environment.)
+`HEROKU_APP_NAME`: Used by the FakeAuthConfig module to determine whether an app is a PR build. If this is set along
```
JPrevost, line 46: This still isn't clear that nobody should actually ever set `HEROKU_APP_NAME` because it happens automatically.

Maybe:

`HEROKU_APP_NAME`: This is set automatically in PR builds and is used by the FakeAuthConfig module to determine whether an app is a PR build. This should never be manually set or overridden in any environment as it is not necessary to do that and can only cause bad things to happen.

JPrevost (approved):
Nice. I like the refactor. Approving, but also made a suggestion on a readme update to be extremely clear that nobody should ever set that value while still explaining what it is used for.

README.md
jazairi: Yes, good call. I added a note along those lines.

jazairi:
@JPrevost Thanks for the detailed review!

PR #72, opened 2024-08-02
Create rake task for bulk reloading of SuggestedResource records

matt-bernhardt:
## Developer

This defines a rake task for importing SuggestedResource records from a CSV file, which must be loaded from a remote URL (via either HTTP or HTTPS protocols, to support local environments). The rake task hands off to a new `bulk_replace` method on the Detector::SuggestedResource class, which actually does the deletion and re-import steps.

There are a few side effects to this change, which are discussed in more detail in the commit message.

Of particular note is the drop in code coverage, which seems to happen because this is the first PR to add tests for any rake task? The task defined here is being reported as 100% coverage - but I'm not sure what our practices are with rake testing in general.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-52

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated
NO migrations are included

### Code

added technical debt.

### Documentation

### Testing


app/models/detector/suggested_resource.rb
```diff
@@ -53,5 +53,27 @@ def calculate_fingerprint(old_phrase)
...
+    # @note This method is called by the suggested_resource:reload rake task.
+    #
+    # @param input [CSV::Table] An imported CSV file containing all Suggested Resource records. The CSV file must have
+    #                           at least three headers, named "Title", "URL", and "Phrase".
+    def self.bulk_replace(input)
+      raise ArgumentError.new, 'Tabular CSV is required' unless input.instance_of?(CSV::Table)
+
+      # Need to check what columns exist in input
+      required_headers = %w[Title URL Phrase]
+      missing_headers = required_headers - input.headers
```
jazairi, line 68: It seems like the inputs are case-sensitive, in which case I'd suggest we underscore that in the method documentation above. (The exact values are already listed, but it may be helpful to explicitly note case sensitivity.)

lib/tasks/suggested_resources.rake
```diff
@@ -0,0 +1,28 @@
...
+  task :reload, [:addr] => :environment do |_task, args|
+    raise ArgumentError.new, 'URL is required' unless args.addr.present?
+
+    raise ArgumentError.new, 'Local files are not supported yet' unless URI(args.addr).scheme
+
+    Rails.logger.info('Reloading all Suggested Resource records from CSV')
+
+    url = URI.parse(args.addr)
+
+    raise ArgumentError.new, 'HTTP/HTTPS scheme is required' unless url.scheme.in?(%w[http https])
```
jazairi, line 17: Does it make sense to only accept HTTPS? (I'd hope that any CSV we're loading is coming through HTTPS and not HTTP, but the Internet is weird sometimes.)

jazairi (changes requested):
Nice work on this! I have a couple of questions and suggestions to improve the documentation, but overall this is looking great. I especially appreciate the robust test coverage for the rake task. I think we normally test those indirectly, but reviewing this made me reconsider whether that's the right approach. I agree with your suggestion to make that a topic for team discussion.

Speaking of which, I should note that I haven't confirmed functionality locally. The coverage is so thorough that I'm not too worried about it. I can do that before approving, though, if you think it would be prudent.

app/models/detector/suggested_resource.rb
matt-bernhardt: Yeah, good point - case sensitivity did bite me initially with this (the File path silently downcases the header values, while URI doesn't (among other changes). I'll add this to the docs.

test/tasks/suggested_resource_rake_test.rb
```diff
@@ -0,0 +1,68 @@
...
+  def setup
+    Tacos::Application.load_tasks if Rake::Task.tasks.empty?
+    Rake::Task['suggested_resources:reload'].reenable
+  end
+
+  test 'reload can accept a url' do
+    records_before = Detector::SuggestedResource.count # We have three fixtures at the moment
+    first_record_before = Detector::SuggestedResource.first
+    VCR.use_cassette('suggested_resource:reload from remote csv') do
+      remote_file = 'http://static.lndo.site/suggested_resources.csv'
```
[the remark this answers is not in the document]
matt-bernhardt: No problem - I can add that.

lib/tasks/suggested_resources.rake
jazairi: Oh, good call. I wasn't thinking about local. I don't see it as a huge issue, fwiw -- no need to make a change.

matt-bernhardt:
Okay, @jazairi - I've added a note to the parameter for the rake task to clarify that the column names are case sensitive, and also added a section to the readme for regenerating the cassettes (including the `.lando.yml` file I'm using to serve these files locally).

jazairi (approved):
Looks great -- thanks!

PR #77, opened 2024-08-05
Add CanCanCan for authorization

jazairi:

We will need authorization for TACOS, as it will contain a variety of features that should not be avilable to every user.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-31

This adds the CanCanCan gem, which [we agreed to use for authorization'](https://github.com/MITLibraries/tacos/blob/main/docs/architecture-decisions/0006-use-cancancan-for-authorization.md). It also introduces a very rudimentary Ability model -- mostly a placeholder until we know what we need -- and a boolean `admin` field to the User model.

It does not implement any rules in the Ability model beyond 'admins can manage anything.'' I debated whether to go even that far, as there's nothing yet to manage.

It's also worth noting that while I was initially opposed to the admin boolean, it does appear to be an idiom within CanCanCan, so I'm comfortable using it here. I think things become more complicated when we combine that pattern with role-based authorization (more on that in side effects, below).

Unless we implement [role-based authorization](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/role_based_authorization.md), as we've done in ETD, CanCanCan seems to want us to include all of our rules in the `initialize` method. That feels like it would get out of hand quickly.

We had briefly discussed in a team meeting that we _don't_ want to implement authorization similarly to ETD, with increasing permissions at each role. (Though, that type of cascading is where [CanCanCan excels](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_check_abilities.md).) It is also possible to assign multiple roles to a single user, [albeit kludgily](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/role_based_authorization.md#many-roles-per-user).

I do think roles are the way to go here, but we'll want to proceed with caution. As mentioned in the ADR, we may learn early on that CanCanCan is not actually the best authorization gem; finding that we need a dozen or more roles may be one trigger to consider an alternative gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,9 @@ end
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', require: false
 
+# Use CanCanCan for authorization
+gem 'cancancan'
+
 # Use Devise for authentication
 gem 'devise'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,6 +100,7 @@ GEM
     bootsnap (1.18.3)
       msgpack (~> 1.2)
     builder (3.3.0)
+    cancancan (3.6.1)
     capybara (3.40.0)
       addressable
       matrix
@@ -430,6 +431,7 @@ PLATFORMS
 DEPENDENCIES
   annotate
   bootsnap
+  cancancan
   capybara
   climate_control
   debug
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   helper Mitlibraries::Theme::Engine.helpers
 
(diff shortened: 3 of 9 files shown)


matt-bernhardt (approved):
This looks good to me - scoped well, simply does what you're intending.

As far as the discussion of side effects, I'm admit to being curious how this is going to grow. From my persual of the documentation you've linked, the only bit that gives me pause is the use of bitmasks for assigning many roles per user. That's not a deal-breaker for me, though - just something that I note with curiosity.

:shipit: is my vote here. Thanks!

jazairi:
@matt-bernhardt Thanks! Yeah, I share your concern about using bitmasks. If multiple roles per user ends up making sense for us, I think we should add a Role model and a join table to set up a has_many relationship. But, that's a discussion for another day.

PR #78, opened 2024-08-07
Add administrate gem ...

matt-bernhardt:
This adds the Administrate gem, and uses the built-in generator for dashboards for four existing resources:

* Detector::SuggestedResources
* SearchEvents
* Terms
* Users

Notable changes to the generator are:
1. The admin application generator gets a `require_user` method in addition to `authorize_user`, to ensure relevant error messages in all cases (confirmed with tests)
2. The Suggested Resources edit form has the fingerprint field removed, because that is a calculated field - so the user shouldn't have the opportunity to change it.

There is also a conditional link in the navigation, shown only to users who have the ability to access the default dashboard (currently set to the Terms area, but will change in a future ticket). This is a divergence from the pattern we're using in ETD, where the nav link is conditional on the admin flag on the User model (which I think is a pattern to avoid).

In hindsight, I think this PR would have been cleaner to have one commit of the generator output, and then a second commit to make the customization changes. If the reviewer would rather see this structure, let me know and I'll restructure accordingly.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-51

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

YES dependencies are updated
NO migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -1,0 +1,36 @@
+# All Administrate controllers inherit from this
+# `Administrate::ApplicationController`, making it the ideal place to put
+# authentication logic or other before_actions.
+#
+# If you want to add pagination or other controller-level concerns,
+# you're free to overwrite the RESTful controller actions.
+module Admin
+  class ApplicationController < Administrate::ApplicationController
+    before_action :require_user
+    before_action :authorize_user
+
+    private
+
+    def authorize_user
+      return if authorize_action?(resource_name, action_name)
+      
+      redirect_to root_path, alert: 'Not authorized'
+    end
+
+    def authorize_action?(resource, action)
+      can? action, resource
+    end
+
+    def require_user
+      return if current_user
+      
+      redirect_to root_path, alert: 'Please sign in to continue'
+    end
+
+    # Override this value to specify the number of elements to display at a time
+    # on index pages. Defaults to 20.
+    # def records_per_page
+    #   params[:per_page] || 20
+    # end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -111,3 +111,5 @@ group :test do
   gem 'vcr'
   gem 'webmock'
 end
+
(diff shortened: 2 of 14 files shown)


app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def require_user
+      return if current_user
+      
+      redirect_to root_path, alert: 'Please sign in to continue'
+    end
+
+    # Override this value to specify the number of elements to display at a time
+    # on index pages. Defaults to 20.
+    # def records_per_page
```
jazairi, line 32: I assume this method was generated by the administrate install? I'm fine keeping it in; just want to make sure it wasn't a leftover comment from a WIP.

jazairi (approved):
Thanks for setting this all up! Everything looks right to me.

This also feels like a good place to share something I learned while researching the issue we were having yesterday with the 'lobby' view. It seems likely that the issue is not with administrate, but with using CanCanCan without a model.  If I'm correct about that, then we need to do two things to get authorization to work properly with the Stat/StatDashboard resource, [as suggested here](https://stackoverflow.com/a/69729278):

* In the controller, specify that there is no associated model: `authorize_resource class: false`
* In the Ability model, name the resource as a symbol: `can :read, :stat`

I tried the above yesterday and it didn't work for me. It was at the end of the day, though, so it might be worth digging into with fresh eyes. (I suspect I was doing something wrong with pluralization.)

app/controllers/admin/application_controller.rb
matt-bernhardt: Yeah, this was what the generator put in place - I remember noting it while working, and thinking "huh, I wonder if we'll ever actually use that".

matt-bernhardt:
Ooh, interesting! Thanks for finding this. I've added it as a reference in the follow-up ticket: https://mitlibraries.atlassian.net/browse/TCO-63

PR #79, opened 2024-08-09
Adds PR template

matt-bernhardt:
This adds a PR template file for this application, based on our other Heroku applications. One notable difference is that the Documentation section makes specific reference to checking the documentation built by Yard.

### Relevant ticket(s)

* https://mitlibraries.atlassian.net/browse/TCO-49
* https://mitlibraries.atlassian.net/browse/ENGX-226


jazairi (approved):
Looks great, and thanks for doing this! I keep thinking about it every time I open a PR, but apparently not enough to actually make the template. :)

PR #84, opened 2024-08-13
Extend Metrics::Algorithms feature to include Detector::SuggestedResource matches

matt-bernhardt:
This extends the `Metrics::Algorithms` model which generates historical records of the application's performance to include matches on the SuggestedResources records that have been added recently. A part of this work is that the `Detector::SuggestedResource` model now gets a `full_term_match` method, which is the crucial link to allow the model to do anything about search traffic.

Along the way the `calculate_fingerprint` method becomes a class method, rather than an instance method. It is still part of the SuggestedResource model, and not in a helper.

Much of this approach has been copied from the Journal title matching work - which seems like a reasonable starting point, both because the uses are similar and because consistent approaches in the app feels like Doing This Right.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-25

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

YES migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/detector/suggested_resource.rb
+++ b/app/models/detector/suggested_resource.rb
@@ -23,12 +23,12 @@ module Detector
     before_save :update_fingerprint
 
     def update_fingerprint
-      self.fingerprint = calculate_fingerprint(phrase)
+      self.fingerprint = Detector::SuggestedResource.calculate_fingerprint(phrase)
     end
 
     # This implements the OpenRefine fingerprinting algorithm. See
     # https://openrefine.org/docs/technical-reference/clustering-in-depth#fingerprint
-    def calculate_fingerprint(old_phrase)
+    def self.calculate_fingerprint(old_phrase)
       modified_phrase = old_phrase
       modified_phrase = modified_phrase.strip
       modified_phrase = modified_phrase.downcase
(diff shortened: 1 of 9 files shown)


app/models/detector/suggested_resource.rb
```diff
@@ -76,5 +76,9 @@ def self.bulk_replace(input)
         record.save
       end
     end
+
```
JPrevost, line 79: Would you mind adding docs for this method and the others that don't have them in this class?

See https://github.com/MITLibraries/tacos/blob/main/app/models/detector/journal.rb#L19-L32 for syntax examples that work well with our documentation server you can run via `make docserver`

app/models/metrics/algorithms.rb
```diff
@@ -107,5 +109,11 @@ def process_journals(event, matches)
       matches[:journal_exact] += 1 if journal_exact.count.positive?
       journal_exact
     end
+
```
JPrevost, line 112: Please add docs. See https://github.com/MITLibraries/tacos/blob/main/app/models/metrics/algorithms.rb#L95-L110 for example syntax and check with `make docserver`

JPrevost (changes requested):
Looks good. Please add the method documentation we agreed to experiment with on this repository.

matt-bernhardt:
Oh, dagnabit. Yeah, I'll add that. Thanks for the prod.

JPrevost (approved):

PR #101, opened 2024-09-10
Implement Detector, Category, DetectorCategory

matt-bernhardt:
This adds models for Detector, Category, and the linking DetectorCategory. There are uniqueness and dependency constraints specified in the migration and class files, and confirmed via tests.

I did not write tests for the linking class, as I'm not sure there's a need? I'm happy to write these if desired, though.

The fixtures for these classes match what we've already implemented in code, with the thought that these should change slowly enough that we should probably keep the fixtures in sync? I'm not sure about this though.

The biggest side effect is that implementing the Detector class is done by changing the existing module to a class - which forces changes in a handful of other files.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-82

### Accessibility

### Documentation

- [x] No documentation changes are needed **These classes will eventually need yard documentation, but thus far I don't know that this is needed?**

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

YES migrations are included **Adding three models with associated records**

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/detector.rb
+++ b/app/models/detector.rb
@@ -6,7 +6,6 @@
 #
 #  id         :integer          not null, primary key
 #  name       :string
-#  confidence :float
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #
--- a/app/models/detector.rb
+++ b/app/models/detector.rb
@@ -1,9 +1,16 @@
 # frozen_string_literal: true
 
-# Detectors are classes that implement various algorithms that allow us to identify patterns
-# within search terms.
-module Detector
-  def self.table_name_prefix
-    'detector_'
-  end
+# == Schema Information
+#
+# Table name: detectors
+#
+#  id         :integer          not null, primary key
+#  name       :string
+#  confidence :float
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Detector < ApplicationRecord
+  has_many :detector_categories, dependent: :destroy
+  has_many :categories, through: :detector_categories
 end
--- a/db/migrate/20240909183413_create_categories.rb
+++ b/db/migrate/20240909183413_create_categories.rb
@@ -1,13 +1,5 @@
 class CreateCategories < ActiveRecord::Migration[7.1]
   def change
-    create_table :detectors do |t|
-      t.string :name
-      t.float :confidence
-
-      t.timestamps
-    end
-    add_index :detectors, :name, unique: true
-
     create_table :categories do |t|
       t.string :name
       t.text :description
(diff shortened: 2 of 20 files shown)


matt-bernhardt:
Quick postscript about this PR - Please check the structure of the migration, as I'm not sure I got this right. With uniqueness constraints on the database tables, and sample records in the migration, I may have created a PR where the rollback doesn't happen cleanly. I've tried working with `up` and `down` methods in the migration, but haven't had success there, so everything is just done in `change` at the moment.

I'm happy to have guidance or suggestions for changes here.

db/migrate/20240909183413_create_categories.rb
```diff
@@ -0,0 +1,44 @@
...
+  def change
+    create_table :detectors do |t|
+      t.string :name
+      t.float :confidence
+
+      t.timestamps
+    end
+    add_index :detectors, :name, unique: true
+
+    create_table :categories do |t|
```
JPrevost, line 11: Non-blocking: There's arguably a benefit to having each of these table creations be in its own migration file to allow for slightly easier context.

db/migrate/20240909183413_create_categories.rb
```diff
@@ -0,0 +1,44 @@
...
+      t.string :name
+      t.text :description
+
+      t.timestamps
+    end
+    add_index :categories, :name, unique: true
+
+    create_table :detector_categories do |t|
```
JPrevost, line 19: we likely want an index that is on the combined ids of the two tables this joins to prevent tables scans. Possibly in both directions depending on our future expected needs (do we need all categories for a specific detector only, or are the times when we want all detectors for a specific category?)

test/models/category_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    record = categories('transactional')
+    link_category = record.detector_categories.count
+
+    record.destroy
+
+    assert_equal(category_count - 1, Category.count)
+    assert_equal(link_count - link_category, DetectorCategory.count)
+  end
+
+  test 'destroying a Category will not delete associated Detectors' do
```
JPrevost, line 38: I came looking for this test while reading the model code :) 👍🏻

app/models/detector.rb
```diff
@@ -1,9 +1,16 @@
...
-  def self.table_name_prefix
-    'detector_'
-  end
+# == Schema Information
+#
+# Table name: detectors
+#
+#  id         :integer          not null, primary key
+#  name       :string
+#  confidence :float
```
JPrevost, line 9: I feel waffly about whether we actually were able to state what `confidence` would even mean at the `Detector` level. If we don't know what it's for, I'd like to leave it out until we need it. If I am mistaken and there was a clear plan for what specifically this mean, please remind me...

My recollection is the `detector_category` confidence was what we felt would increment/decrement during the validation process.

JPrevost (commented):
I need to pivot to something else and will finish the review tomorrow, but wanted to submit my initial input even if it isn't complete. Overall this looks like what we want.

I definitely think the db/seed.rb is the better way to handle the data creation you have in the migration (whether we seed it in prod or manually create it is tbd, but it'll be great to have that seed to auto run them in PR builds and locally whenever we find a need).

matt-bernhardt:
Thanks - let me look over these suggestions and make some changes. I'll un-tag for a review now, and re-tag when I'm done.

db/migrate/20240909183413_create_categories.rb
```diff
@@ -0,0 +1,44 @@
...
+      t.belongs_to :detector, null: false, foreign_key: true
+      t.belongs_to :category, null: false, foreign_key: true
+      t.float :confidence
+
+      t.timestamps
+    end
+
+    Detector.create(name: 'DOI', confidence: 0.95)
```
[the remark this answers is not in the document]
matt-bernhardt: Yeah, that makes sense. I forgot that we use that approach for licenses and whatnot in ETD. I'll change this.

db/migrate/20240909183413_create_categories.rb
matt-bernhardt: I'm happy to do whatever is most expected in Rails. The migrations were originally separate files, which I combined thinking that this was one feature and set of related tables. But, that did introduce some uncertainty about the name of the migration (it is inaccurately named at the moment).

db/migrate/20240909183413_create_categories.rb
matt-bernhardt: Is there a cost to building both right now, just in case? I'm not clear right now what we'd need. During categorization, starting from a given term, it would probably be "take the detectors for which this term activated, and increment their associated category scores" - which feels like a bit like "find the categories associated with this detector". For the final population of category scores, though direction might be the other way - I'm not sure yet.

app/models/detector.rb
matt-bernhardt: I don't think we ever finished that conversation. I saw you were thinking about just going with a confidence value on the link record, but Adam hadn't weighed in yet. I'm not sure - it might be needed, but I'm happy to go without it until we realize that for sure. The initial walkthrough seemed to indicate that we might not need a detector confidence.

I'll take it out for now.

db/migrate/20240909183413_create_categories.rb
JPrevost: Minimal cost for having both. (In fact the two tables will both be so small that even not having these indexes would be fine... but I didn't say that!)

matt-bernhardt:
Okay, @JPrevost - I've made the changes you mentioned in your initial review. It should be ready for you to finish looking at now.

db/seeds.rb
```diff
@@ -7,3 +7,56 @@
 #   ["Action", "Comedy", "Drama", "Horror"].each do |genre_name|
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
+
```
JPrevost, line 10: Note: If these don't run automatically in our PR builds (and maybe other Heroku tiers? We can talk about that) they can be configured to do that. I haven't checked if this repo is already set to run seeds in Heroku or not, just noting it before I forget as something we can address if it doesn't work as desired.

PR #107, opened 2024-09-18
Implement Detections model

matt-bernhardt:
This defines the `Detections` model, which links our Term and Detector records. The details have gotten a bit messier than I expected, and some changes to the class diagram may be necessary. This PR updates that diagram to help make it clear what's changing:

First, the confidence value is no longer stored as part of the Detections record. Because a Detector might be mapped to multiple Categories, each with its own confidence value, but the Detections record has no relationship to Category, there's no way to comprehensively cache the appropriate confidence value here. The values will still be used during the Categorization process, but we'll have to look them up in the graph rather than cache them with the Detection.

Second, the various record_* methods are defined in the Term model, and not in the Detection model.

I've tried to write thorough tests confirming the various constraints, and walking through the intended uses of this model. I'm not sure that I've gotten everything, but hopefully nothing is glaringly wrong.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-33

### Accessibility

### Documentation

### ENV

- [x] All new ENV is documented in README.
- [ ] ENV has not changed.

### Stakeholders

NO dependencies are updated

**YES migrations are included**

### Code

added technical debt.

### Documentation

### Testing


app/models/detection.rb
```diff
@@ -0,0 +1,52 @@
...
+  scope :for_detector, ->(detector) { where(detector_id: detector.id) }
+  scope :for_term, ->(term) { where(term_id: term.id) }
+
+  private
+
+  # This leverages three scopes to confirm, before creating a new Detection, whether that combination of Term, Detector,
+  # and Version has already been registered. If one is found, it aborts the creation.
+  def check_for_current
+    return unless Detection.for_term(term).for_detector(detector).current.exists?
+
```
JPrevost, line 41: I'm trying to understand a use case where this logic is preferred over using a `Detection.find_or_create_by` to always return a `Detection` (either an existing one ore creating and returning the new one).

app/models/term.rb
```diff
@@ -11,4 +14,64 @@
...
+  # Detector::StandardIdentifiers. Any matches will be registered as Detection records.
+  #
+  # @note There are multiple checks within the Detector::StandardIdentifier class. Each check is capable of generating
+  #       a separate Detection record (although a single check finding multiple matches would still only result in one
+  #       Detection for that check).
+  def record_patterns
+    si = Detector::StandardIdentifiers.new(phrase)
+
+    si.identifiers.each_key do |k|
+      Detection.create(
```
JPrevost, line 38: This (and the similar blocks) might be where we use `find_or_create_by` rather than the custom logic currently in `Detector`. Thoughts?

app/models/detection.rb
matt-bernhardt: That's an interesting thought, and it would allow us to avoid this method. The biggest drawback I can come up with is that `find_or_create_by` actually returns a record - and these methods don't actually do anything with that - they don't return anything, and take no actions after creating the records.

Is that enough of a reason not to use `find_or_create_by`? I'm not sure. If you'd rather go that route, I'm happy to give it a shot. We could probably ditch all three scopes and this private method (the other one would still be needed, though).

matt-bernhardt: The more I think about this, the more I think it makes sense to use `find_or_create_by` in this context. I think there might still be a need for the scopes, but that's just a suspicion rather than anything we've discussed clearly.

I'll take this back out of review and push an updated commit soon.

matt-bernhardt:
Okay, @JPrevost - I think this is ready for another review. I've elected to keep the scopes in place - I think there's going to be a clear need for the `current` and `for_term` scopes during categorization. I'm less certain about the `for_detector` scope, but the symmetry with `for_term` makes a certain sense to me at this point.

app/models/detection.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# A Detection is a joining record between a Term and a Detector, created when a Detector activates based on some aspect
+# of the Term. This is the record that TACOS found something about this Term.
+#
+# There is a uniqueness constraint on the combination of term_id, detector_id, and detector_version.
+#
+# New records can be created by passing a Term and a Detector object. The model will look up the current detector
+# version, and include that in the record. If this would result in a duplicate record, then the creation process is
+# aborted with no duplicate created.
```
JPrevost, line 10: Minor suggestion. Rather than saying the process is aborted, I'd say it will throw an exception. One implies maybe we silently continue (but I don't believe we catch the exception to recover cleanly) whereas the other is clear we'll blow up because this should never happen kinda thing. That said, I'm not sure we need this note. I'm not against it, but it feels like something that may easily get outdated as it describes how the model behaves rather than just what it is for. 🤔

app/models/detector.rb
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+# Detectors are a generic representation of specific tests performed by the
```
JPrevost, line 3: I think `specific tests` as a phrase may lead to confusion when working in a context in which we have test suites that mean something else. Maybe `specific checks` or similar?

test/models/detection_test.rb
```diff
@@ -0,0 +1,93 @@
...
+    assert_raises(ActiveRecord::RecordNotUnique) do
+      Detection.create!(new_sample)
+    end
+
+    # ...but when we update the DETECTOR_VERSION env, now the same record does save.
+    new_version = 'updated'
+
+    assert_not_equal(ENV.fetch('DETECTOR_VERSION'), new_version)
+
+    ENV['DETECTOR_VERSION'] = new_version
```
JPrevost, line 60: Setting env in line like this may lead to instability in the test suite. We often use the ClimateControl gem to help mitigate those risks in our tests. I'm not sure if we've used that yet in this application, but if we need to change env in the test suite I think it ensures it gets set back after the test in a safer way.

JPrevost (approved):
Overall I believe this works. I have two comments (one small and one potential refactor). Neither are blocking so I am approving. If you choose to make any changes please dismiss my review when you are ready for a re-review.

app/models/term.rb
```diff
@@ -11,4 +14,64 @@
...
+    record_suggested_resources
+  end
+
+  # The record_patterns method will consult the set of regex-based detectors that are defined in
+  # Detector::StandardIdentifiers. Any matches will be registered as Detection records.
+  #
+  # @note There are multiple checks within the Detector::StandardIdentifier class. Each check is capable of generating
+  #       a separate Detection record (although a single check finding multiple matches would still only result in one
+  #       Detection for that check).
+  def record_patterns
```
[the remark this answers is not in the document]
matt-bernhardt: I'm going to think about this idea tomorrow morning - I think I like it, but I'm also not sure whether it results in code spread across the application for one coherent work step. The prospect of the Term model not needing to keep tabs on everything is an attractive one, however.

JPrevost (approved):
Looks great. `record` will likely always hurt my brain as I'm also thinking of that as a noun and not a verb. Hopefully we get used to it and if not we can change it.

PR #119, opened 2024-10-09
Add a citation detector

matt-bernhardt:

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-96

### Accessibility

### Documentation

### ENV

- [ ] ENV has not changed.

### Stakeholders

- [ ] Stakeholder approval is not needed

YES | NO dependencies are updated

YES | NO migrations are included

### Code

added technical debt.

### Documentation

### Testing


matt-bernhardt:
Oh, @JPrevost - a question that occurs to me now, but which may just complicate things: even though this citation detector doesn't provide any information to the GraphQL endpoint, should there be a test in the GraphQL controller to confirm that traffic coming in from the API does actually result in this Detector being tripped? That feels like an integration test, rather than a test of the GraphQL endpoint itself?

I'm happy to write up a test suite like this if you'd like.

app/models/detector/citation.rb
```diff
@@ -0,0 +1,156 @@
...
+
+    # The initializer handles the parsing of a Term object, and subsequent population of the @subpatterns, @summary,
+    # and @score instance variables. @subpatterns contains all the citation components which have been flagged by the
+    # CITATION_PATTERNS hash. @summary contains counts of how often certain characters or words appear in the Term.
+    # Finally, the @score value is a summary of how many elements in the subpatterns or summary report were detected.
+    #
+    # @note This method can be called directly via Detector::Citation.new(Term). It is also called indirectly via the
+    #       Detector::Citation.record(Term) instance method. This method can be called directly when a Detection is not
+    #       desired.
+    def initialize(term)
```
JPrevost, line 52: I've been thinking about a future state of having tacos run in "not logging mode". Having our detectors work based on `Strings` instead of `Terms` would be key to that potential functionality.

The `.record` methods could still use the `Term` as those are explicitly part of our write mode processes.

I'm sharing this just as information on some future direction we may head. If you prefer to leave this initialize method using the Term that is fine. Some of our Detectors use Terms and some use Strings. Eventually I want to normalize that as it is a type of technical debt to have them work differently for now good reason. As we haven't formally picked a path forward, either Term is String is acceptable, but I'm leaning towards proposing we adopt Strings for all Detectors even if we use Terms for `record` to give us the flexibility checking how TACOS would respond to input without having to store that input. The main use cases I have for that desired feature is using the playground without having to store my test queries as Terms as well as potentially opening up TACOS for others to experiment with without us introducing a privacy concern storing data from other systems not in our control. The latter is not planned so the former is what I'd like to decide how to handle.

app/models/detector/citation.rb
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+    # This builds one of the two main components of the Citation detector - the summary report. It calls each of the
+    # methods in the first line - which all return integers - and puts the result as a key-value pair in the @summary
+    # instance variable.
+    #
+    # @return hash
+    def summarize(term)
+      %w[characters colons commas periods semicolons words].each do |check|
+        @summary[check.to_sym] = send(check, term)
```
JPrevost, line 147: This took me a bit to follow what was going on even with the docs. It's a solid pattern, but we don't reach for it enough for me to have immediately understood it.

app/models/detector/citation.rb
```diff
@@ -0,0 +1,156 @@
...
+    def initialize(term)
+      @subpatterns = {}
+      @summary = {}
+      pattern_checker(term.phrase)
+      summarize(term.phrase)
+      @score = calculate_score
+    end
+
+    # The record method first runs all of the parsers by running the initialize method. If the resulting score is higher
+    # than the REQUIRED_SCORE value, then a Detection is registered.
```
JPrevost, line 61: It feels a bit odd to me that the record method is where the decision as to whether the input is a citation or not is made.

I wonder if adding a boolean method "citation?" would be valuable. That would give us more flexibility to report that information to UIs that might want it in the future (I'm fine with not returning anything in GraphQL just yet as we learn more about this Detector's usefulness, but having something in GraphQL that indicated we think it was a citation does feel potentially useful)

```ruby
def citation?
  @score >= REQUIRED_SCORE
end
```

Then record could leverage that rather than doing that calculation itself.

```ruby
return unless cit.citation?
```

JPrevost (commented):
Overall this looks really good and is very interesting.

I added a couple of comments that I'd like to let you decide how to address (or not) before finalizing the review as they could potentially change enough that I'd want to start my review over. Neither are blocking, so if you want to leave them as-is please let me know and I'll complete the review with the code as-is.

app/models/detector/citation.rb
matt-bernhardt: Yeah, this feels like a good method. I'll revise the PR to add this, at least. What do you think about `detection?` as a name, though? That might be shareable across models?

app/models/detector/citation.rb
matt-bernhardt: Is there a better way to say "call all these methods" ? I was looking for a structure that was efficient without being opaque, so I'm open to hearing that this doesn't quite hit that spot.

JPrevost: No, I think what you are doing here is great. `send` is meant for this. I just had to look it up as we don't use it much.

app/models/detector/citation.rb
JPrevost: `@detection` is what I'm toying with for all pattern matches detected as an instance variable. The expectation would be that it would be an array of name/value pairs.

Having a boolean `detection?` as a convenience method (which on some models would just be the presence of anything in `detection` but for this could be a subset that match the score threshold) does seem like a good pattern.

app/models/detector/citation.rb
matt-bernhardt: I support having a path through the application to allow for querying with strings, but I'd prefer to make that change in a coordinated way. Of the detectors we have now, every example that uses an `initialize` method (StandardIdentifiers, LCSH, and now Citation) does so based on a Term method - so I'd prefer to keep this consistency. I worry about having a state, even temporarily, where one detector behaves differently from the others.

As part of such a change, I wonder about how to name the methods to clarify what is expected, i.e. something like:

```
Detector::Citation.new(phrase) # using built-in "new" as a method

Detector::Citation.register_term(term) # method named to clarify expected argument, instead of "record"
```

But, we can have this discussion more thoroughly in whatever ticket centers that refactoring.

app/models/detector/citation.rb
matt-bernhardt: I've pushed a commit and tests to add this, and converted the `record` method to consume it.

matt-bernhardt:
Okay, I think this is ready for review - I've added a commit to implement a convenience method for the activation check. I'd prefer to handle the change from Terms to Strings as input in a coordinated way, leaving the Detectors to behave consistently for now.

app/models/detector/citation.rb
JPrevost: Yup, that makes sense to me.

JPrevost (approved):
Please remember to create the Detector record manually in Stage/Prod when promoting this work.

:shipit:

PR #126, opened 2024-10-17
Integrate Citation detector with Metrics

matt-bernhardt:
This integrates the citation detector with the monthly metrics setup. It includes a tweak to one of the term fixtures, because the ISBN fixture had ended up getting counted in the citation detector (which is a real phenomenon, but our test fixtures feel like they should be crafted to keep things separate).

### Quick note about the side effect

The fact that this needed to tweak a different term fixture is an indication that our detectors are starting to overlap. This is a real phenomenon in production, and is expected as we start to build more detectors.

I do not have a notion for now about how to address this in the more constrained test environment, where we can carefully craft fixtures to do exactly what we intend. It feels like this behavior should be confirmed via explicit tests that address this overlap, but the tests we have so far are focused on one-and-only-one detector, so I feel okay about tweaking things to keep them separate.

We do have a term fixture that trips multiple detectors, confirming that they get counted in all applicable categories, so we may be covered already?

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-107

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

YES migrations are included

### Code

- [x] I have confirmed that the code works as intended.
added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/metrics/algorithms.rb
+++ b/app/models/metrics/algorithms.rb
@@ -48,8 +48,9 @@ module Metrics
                 else
                   count_matches(SearchEvent.includes(:term))
                 end
-      Metrics::Algorithms.create(month:, doi: matches[:doi], issn: matches[:issn], isbn: matches[:isbn],
-                                 lcsh: matches[:lcsh], pmid: matches[:pmid], journal_exact: matches[:journal_exact],
+      Metrics::Algorithms.create(month:, citation: matches[:citation], doi: matches[:doi], issn: matches[:issn],
+                                 isbn: matches[:isbn], lcsh: matches[:lcsh], pmid: matches[:pmid],
+                                 journal_exact: matches[:journal_exact],
                                  suggested_resource_exact: matches[:suggested_resource_exact],
                                  unmatched: matches[:unmatched])
     end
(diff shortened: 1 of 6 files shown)


app/models/metrics/algorithms.rb
```diff
@@ -78,11 +79,25 @@ def count_matches(events)
     # @return does not return anything (the same matches Hash is passed in each loop but not explicitly sent back)
     def event_matches(event, matches)
       ids = match_standard_identifiers(event, matches)
+      citation = match_citations(event, matches)
       journal_exact = process_journals(event, matches)
       suggested_resource_exact = process_suggested_resources(event, matches)
       lcshs = match_lcsh(event, matches)
 
-      matches[:unmatched] += 1 if ids.detections.blank? && lcshs.detections.blank? && journal_exact.count.zero? && suggested_resource_exact.count.zero?
+      matches[:unmatched] += 1 if ids.detections.blank? && !citation.detection? && lcshs.detections.blank? && journal_exact.count.zero? && suggested_resource_exact.count.zero?
```
JPrevost, line 87: I'd like to normalize `Citation` a bit more after seeing how it behaves differently. Not in this PR, but just eventually. This could take the form of adding the `.detection?` to all of our Detectors and/or normalizing what `.detection` returns. No changes are should be made as part of this work, but this difference made me pause and have to understand _why_ it was done this way. It makes sense based on how Citation currently works but feels like it may be confusing in the future as to why it is different.

JPrevost (approved):

app/models/metrics/algorithms.rb
matt-bernhardt: Yeah, I agree that having a more standardized set of methods across our detectors would be helpful. Some differences might be unavoidable (the citation detector doesn't extract specific patterns, while the standard identifier one does), but looking now I think the others (which all treat the phrase holistically) could benefit from a `.detection?` method.

PR #138, opened 2024-11-13
Add automatic fingerprinting for Term records

matt-bernhardt:
This adds a TermFingerprint model, which allows for clustering of terms by a shared fingerprint. Full details are in the commit message.

### Ticket(s)

This started as an experiment that kept seeming promising, so I didn't create a ticket for it. Now that I'm writing the PR text, I think maybe I missed an opportunity to do this - but am not going to paper over that decision now without checking in with the project team.

https://mitlibraries.atlassian.net/browse/TCO-124

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

YES migrations are included

### Code

added technical debt.

### Documentation

### Testing


test/models/term_fingerprint_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    target_term = target.terms.last
+
+    assert_operator 0, :<, target.terms.count
+
+    target.destroy
+    target_term.reload
+
+    assert_equal term_count, Term.count
+    assert_equal fingerprint_count - 1, TermFingerprint.count
+    assert_instance_of NilClass, target_term.term_fingerprint
```
JPrevost, line 37: You may want to consider `assert_nil` here:
https://ruby-doc.org/3.0.5/gems/minitest/Minitest/Assertions.html#method-i-assert_nil

Both work, but I think the more explicit assertion is slightly more efficient to mentally process (for me at least)

app/models/term.rb
```diff
@@ -67,6 +86,36 @@ def calculate_categorizations
...
+  # The store_fingerprint method gets called before a Term record is saved, ensuring that Terms should always have a
+  # related TermFingerprint method.
+  def store_fingerprint
+    self.term_fingerprint = TermFingerprint.find_or_create_by({ fingerprint: calculate_fingerprint })
+  end
+
+  # This is similar to the SuggestedResource fingerprint method, with the exception that it also replaces &quot; with "
+  # during its operation. This switch may also need to be added to the SuggestedResource method, at which point they can
+  # be abstracted to a helper method.
+  def calculate_fingerprint
```
JPrevost, line 98: I don't see any tests that confirm this logic works as expected.

You can likely snag them from https://github.com/MITLibraries/tacos/blob/main/test/models/detector/suggested_resource_test.rb
as we'll be refactoring and removing those tests soon to use the Term fingerprints.

This feels too important to hide behind "we don't test private methods" to me. If you don't test it directly (which is fine for a private method), we still should test the effects of it by testing the generated fingerprints on a Term (although it's probably slightly easier to just test it directly).

I'm also curious if this might better belong in the Fingerprint model itself. It feels slightly odd to have a model dedicated to fingerprints but have the actual logic in Term.

app/models/term_fingerprint.rb
```diff
@@ -0,0 +1,16 @@
...
+# == Schema Information
+#
+# Table name: term_fingerprints
+#
+#  id          :integer          not null, primary key
+#  fingerprint :string
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+class TermFingerprint < ApplicationRecord
```
JPrevost, line 12: Now that we have decided to only have one set of stored fingerprints, it feels worth considering whether to name this `Fingerprint` instead of `TermFingerprint`. This is non-blocking, but if you choose to not make this change I'd like to hear why you prefer it this way.

As noted elsewhere, I'd prefer if we can move the logic to calculate a fingerprint to this model rather than keep it in Term.

JPrevost (changes requested):
Blocking:

- I'd like to see some unit tests for the fingerprint algorithm.

Suggestions (non-blocking): 
- rename TermFingerprint to Fingerprint
- move fingerprinting logic out of Term and into Fingerprint (I'm nearly blocking on this so please consider it strongly and if you don't think it is the correct path let's chat so I can understand why)
- use `assert_nil`

matt-bernhardt:
My question about moving TermFingerprint to Fingerprint (which makes sense on paper) is whether I should retcon the two migrations in this PR to just use that name, or instead create new migrations to drop this table and create the new one. I suspect that the answer is to retcon, and just re-create the review app for this and advise everyone to rollback / migrate carefully - but that feels like something to confirm with you and Adam.

JPrevost:
@matt-bernhardt I'd go with retcon. Make sure to rollback your local database before doing that or the retconned tables will never go away for you. And yeah, for the PR build just make a new one to avoid any confusion.

matt-bernhardt:
I'm putting this into draft mode while I work on this refactoring.

app/models/term_fingerprint.rb
matt-bernhardt: Do you have thoughts about naming the class versus naming the field? Using the same name for both feels like it would get awkward, particularly from the perspective of the Term model. `term.fingerprint` feels like the method to call to get the calculated string, but if the class is also Fingerprint then things get a bit weird.

For now I'm using `.fingerprint_value` as the Term method, but I'm not sure that's going to be the case when it gets to code review.

app/models/term.rb
matt-bernhardt: Thankfully, ending up with an instance method for this logic means that we can't hide behind "we don't test private methods" anymore. Either way, I agree that tests are needed here - not sure why I didn't write them initially.

I've tried to make clear in the test structure that there's a test for every line in the method, to make sure we keep focus on its components rather than trying to manage one all-encompassing example. Hopefully future-us will continue to keep up if this logic changes.

app/models/term_fingerprint.rb
matt-bernhardt: One thing I'm particularly interested in your feedback about is the naming being used here. Having a `Fingerprint` model, and a `fingerprint` field, feels awkward, particularly from the Term perspective. I've tried to resolve this by adding the `.fingerprint_value` construct that delegates down to the Fingerprint model, and allows the Term to just get access to the string value somehow.

That said, I don't think this gets used anywhere yet - so maybe it would makes sense to remove it until the UI or something leverages it.

matt-bernhardt:
Okay @JPrevost - I think the fingerprinting PR is once again ready for a review. I think I've picked up the things you asked for changes on, and I think things are better as a result. Let me know what you think?

db/migrate/20241211195504_add_fingerprint_to_terms.rb
```diff
@@ -0,0 +1,14 @@
+class AddFingerprintToTerms < ActiveRecord::Migration[7.1]
+  def up
+    add_reference :terms, :fingerprint, foreign_key: true
+
+    # Seed the relationship between Terms and Fingerprints
+    Term.all.each do |t|
```
JPrevost, line 6: This took a very long time and then failed locally for me with a large set of Terms loaded.

I'd like to load the full set of Terms/SearchEvents into stage before we merge to confirm we won't have issues in prod. That is likely blocked by stage db size so I'll pull that ticket to unblock this work.

For now I'll send you the error I saw via Slack and I suggest you load the full set from prod (you have to load in batches) so you can try to replicate the problem I ran into. Loading in batches by month is how I've approached this as we can't export the full set at this time.

JPrevost: Additionally, I want to consider not doing this large data process in a migration (I know, I know) and instead making it a rake task we run after migration. If I understand this work, fingerprint is not required and thus the migration should be fine without it and we can have a task that regenerates all fingerprints to run manually after this work lands both to generate the full set of fingerprints initially but also in case our fingerprint algorithm ever changes this would allow us to regenerate them.

JPrevost (commented):
Non-blocking: I'd like us to consider `Fingerprint.value` instead of `Fingerprint.fingerprint`. It's not great, but it feels slightly less confusing.

Also non-blocking, but consider whether the default state in fixtures of Terms having no Fingerprints feels correct to you.

If you don't intend to make either of those changes just let me know and I'll approve as the code works as needed. Thanks!

app/models/fingerprint.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: fingerprints
+#
+#  id          :integer          not null, primary key
+#  fingerprint :string
```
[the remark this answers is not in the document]
matt-bernhardt: That seems worth trying, yeah - I'll take a run at it, and see if I can make it work.

test/fixtures/terms.yml
```diff
@@ -2,16 +2,20 @@
...
-#  phrase     :string
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
-#  flag       :boolean
+#  id             :integer          not null, primary key
+#  phrase         :string
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  flag           :boolean
+#  fingerprint_id :integer
```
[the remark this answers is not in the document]
matt-bernhardt: I can see this perspective. I made the last round of changes because I wanted to confirm that not populating the fingerprints as part of the migration wouldn't cause a problem, and it looks like that's the case (no code changes were needed, just test changes to adapt to the new starting point).

Let me see if I can find a comfortable test position here...

matt-bernhardt:
@JPrevost I've added two commits this afternoon, changing the `fingerprint` field to `value`, and updating the state of the fixtures. I feel pretty good about both changes - not "this is perfect", but "this is, on balance, better than it was". It's ready for your review.

app/models/term.rb
```diff
@@ -68,6 +87,22 @@ def calculate_categorizations
 
   private
 
+  # register_fingerprint method gets called before a Term record is saved, ensuring that Terms should always have a
+  # related Fingerprint method.
+  def register_fingerprint
+    new_record = {
+      value: Fingerprint.calculate(phrase)
+    }
+    self.fingerprint = Fingerprint.find_or_create_by(new_record)
```
JPrevost, line 96: Non blocking. You may intentionally be making the hash, but I wanted to share it is not necessary and you could instead just pass the key and value directly without the intermediary object creation.

`self.fingerprint = Fingerprint.find_or_create_by(value: Fingerprint.calculate(phrase))`

JPrevost (approved):

app/models/fingerprint.rb
```diff
@@ -0,0 +1,35 @@
...
+#  value      :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Fingerprint < ApplicationRecord
+  has_many :terms, dependent: :nullify
+
+  validates :value, uniqueness: true
+
+  alias_attribute :fingerprint_value, :value
```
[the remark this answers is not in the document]
matt-bernhardt: For now, I think I like the ergonomics of having it directly on the `Term` model. This may turn out to be something we remove later, which would be fine, but I'd rather deal with that down the road at this point.

PR #143, opened 2024-11-20
Adds preprocessor for incoming primo searches

JPrevost:

* Primo has a more complex keyword search capability than tacos is able to understand
* Rather than adding complexity to what tacos understands to match primo, we are adding a normalization step via source preprocessors

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-113

* This adds our first source preprocessor to handle incoming searches from primo ui
* It does not yet handle complex searches (targeted field searching or multiple keywords being combined with boolean logic)
* When a complex search is detected, we attach it to a shared term 'unhandled complex primo query' so we can better understand the volume of these more complex queries. This should provide us with some initial data to help us prioritize handling more complex queries within tacos.

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/preprocessor_primo.rb
+++ b/app/models/preprocessor_primo.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# PreprocessorPrimo handles manipulating incoming data from the Primo UI into a structure that TACOS can work with
+class PreprocessorPrimo
+  # Processes raw incoming query from Primo, looks at each part to see if it is a keyword anywhere search
+  # Any portion that is not a keyword anywhere search drops the entire search from TACOS, logging
+  # as the shared Term `unhandled complex primo query` to allow us to track how frequently we are
+  # dropping terms so we can come back later to build out more complex handing if this is common enough
+  # to warrant the additional work
+  def self.to_tacos(query)
+    # split on agreed upon joiner `;;`
+    split_query = query.split(';;')
+    # if more than one
+    if split_query.count > 1
+      Rails.logger.debug('Multipart primo query detected')
+
+      # parts = []
+      # split_query.each do |q|
+      #   parts << extract_keyword(q)
+      # end
+
+      # drop anything that isn't `any,contains` as TACOS does not process those yet
+      # return 'unhandled complex primo query' if parts.any?('unhandled complex primo query')
+
+      # combine what is left
+      # parts.join(' ')
+      'unhandled complex primo query'
+    else
+      Rails.logger.debug('Simple primo query detected')
+      # confirm it is `any,contains`
+
+      extract_keyword(query)
+    end
+  end
+
+  # confirms whether a portion of a primo query is a keyword search
+  # @param [Array] query_part_array
+  def self.keyword?(query_part_array)
+    return false unless query_part_array.count >= 3
+    return false unless query_part_array[0] == 'any'
+
+    # For now, we are allowing all variants of the second portion of the primo query input
+    # The expected values are: contains, exact, begins_with, equals
+    #
+    # return false unless query_part_array[1] == 'contains'
+
+    true
+  end
+
+  # extract keyword work at the level of a single keyword query input coming from primo and
+  # returns a string with just that keyword with the operators removed
+  # @param [String] query_part
+  # @return [String] the extracted keyword phrase
+  def self.extract_keyword(query_part)
+    query_part_array = query_part.split(',')
+
+    return 'unhandled complex primo query' unless keyword?(query_part_array)
+
+    query_part_array.slice(2..).join(',')
+  end
+end
--- a/app/models/preprocessor_primo.rb
+++ b/app/models/preprocessor_primo.rb
@@ -9,8 +9,8 @@ class PreprocessorPrimo
   # to warrant the additional work.
   # @param query [String] example `any,contains,this is a keyword search`
   def self.to_tacos(query)
-    # split on agreed upon joiner `;;`
-    split_query = query.split(';;')
+    # Primo and TACOS agreed upon joiner is `;;;`
+    split_query = query.split(';;;')
 
     if split_query.count > 1
       Rails.logger.debug('Multipart primo query detected')
--- a/app/models/preprocessor_primo.rb
+++ b/app/models/preprocessor_primo.rb
@@ -49,7 +49,12 @@ class PreprocessorPrimo
   def self.extract_keyword(query_part)
     query_part_array = query_part.split(',')
 
-    return 'invalid primo query' unless query_part_array.count >= 3
+    # We don't anticipate this being a normal state so we are tracking it under the Term `invalid primo query` as well
+    # as sending an exception to Sentry so we can understand the context in which this happens if it does
+    if query_part_array.count < 3
+      Sentry.capture_message('PreprocessorPrimo: Invalid Primo query during keyword extraction')
+      return 'invalid primo query'
+    end
 
     the_keywords = join_keyword_and_drop_extra_parts(query_part_array)
 
--- a/app/models/preprocessor_primo.rb
+++ b/app/models/preprocessor_primo.rb
@@ -2,11 +2,12 @@
 
 # PreprocessorPrimo handles manipulating incoming data from the Primo UI into a structure that TACOS can work with
 class PreprocessorPrimo
-  # Processes raw incoming query from Primo, looks at each part to see if it is a keyword anywhere search
+  # to_tacos processes raw incoming query from Primo, looks at each part to see if it is a keyword anywhere search
   # Any portion that is not a keyword anywhere search drops the entire search from TACOS, logging
   # as the shared Term `unhandled complex primo query` to allow us to track how frequently we are
   # dropping terms so we can come back later to build out more complex handing if this is common enough
-  # to warrant the additional work
+  # to warrant the additional work.
+  # @param query [String] example `any,contains,this is a keyword search`
   def self.to_tacos(query)
     # split on agreed upon joiner `;;`
     split_query = query.split(';;')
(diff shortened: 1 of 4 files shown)


app/models/preprocessor_primo.rb
```diff
@@ -0,0 +1,69 @@
...
+  # @param [String] query_part
+  # @return [String] the extracted keyword phrase
+  def self.extract_keyword(query_part)
+    query_part_array = query_part.split(',')
+
+    return 'invalid primo query' unless query_part_array.count >= 3
+
+    the_keywords = comma_handler(query_part_array)
+
+    return 'unhandled complex primo query' unless keyword?([query_part_array[0], query_part_array[1], the_keywords])
```
matt-bernhardt, line 54: Something about this feels awkward to me. Explicitly composing the array in this way while calling `keyword?`, and then checking there to confirm that there are only three elements, doesn't seem like it accomplishes anything? If `the_keywords` doesn't get parsed correctly, the argument is still only a three-element array - but that third element might not be a single string - maybe `['foo', 'bar', ['baz', 'none']]` and not `['foo', 'bar', 'baz', 'none']`, since `the_keywords` would only ever be one thing?

test/models/preprocessor_primo_test.rb
```diff
@@ -0,0 +1,102 @@
...
+  test 'keyword? returns true for any contains phrase pattern' do
+    input = 'any,contains,popcorn anomoly'.split(',')
+
+    assert(PreprocessorPrimo.keyword?(input))
+  end
+
+  test 'keyword? returns false for input with more than 3 array elements' do
+    # NOTE: this query entering tacos would work... but it would have been cleaned up prior to running
+    # keyword? in our application via the normal flow
+    input = 'any,contains,popcorn anomoly: why life on the moon is complex, and other cat facts'.split(',')
```
matt-bernhardt, line 41: See above comment about the calling of this method never being able to match the input of this test - we're explicitly composing the argument as a three-element array, if I follow the logic above?

matt-bernhardt (changes requested):
Most of my comments here are to confirm my assumptions about the change, but there are two things that I suspect might need a tweak - but I'm open to pushback or hearing that I'm off base.

1. I don't think the check for three elements in the `keyword?` method is doing anything, given how we are calling that method
2. I think the comment in the `comma_handler`method isn't accurate in its example

Everything else seems fine, and I particularly like the structure of the `extract_phrase` method in SearchLogger - that feels nicely extensible for other systems to start contributing, while also managing non-production environments.

My last comment about re-organizing the GraphQL controller test file is non-blocking - if you agree that it seems worth doing, then I'm happy to make a ticket in the backlog for it.

app/models/preprocessor_primo.rb
```diff
@@ -0,0 +1,69 @@
...
+  # Any portion that is not a keyword anywhere search drops the entire search from TACOS, logging
+  # as the shared Term `unhandled complex primo query` to allow us to track how frequently we are
+  # dropping terms so we can come back later to build out more complex handing if this is common enough
+  # to warrant the additional work
+  def self.to_tacos(query)
+    # split on agreed upon joiner `;;`
+    split_query = query.split(';;')
+
+    if split_query.count > 1
+      Rails.logger.debug('Multipart primo query detected')
```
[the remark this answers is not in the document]
JPrevost: They will show up in environments in which the log level is set to DEBUG which is the default in development in this app.

In production we default to INFO, but can swap to DEBUG with an ENV change to see something in a PR build (or even in prod) for a bit:
https://github.com/MITLibraries/tacos/blob/main/config/environments/production.rb#L65

More info on [Rails log levels](https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels)

Some of our apps we have put the ENV control of log level in development for when we have a few too many debug logs for normal use. I believe in at least one app we run prod in debug log level mode. Consistency would probably be nice :)

app/models/preprocessor_primo.rb
```diff
@@ -0,0 +1,69 @@
...
+    return 'unhandled complex primo query' unless keyword?([query_part_array[0], query_part_array[1], the_keywords])
+
+    the_keywords
+  end
+
+  # comma_handler handles the logic necessary to join searches that contain commas into a single ruby string
+  # after we separate the incoming string into an array based on commas
+  def self.comma_handler(query_part_array)
+    # Join the third to the end of the into a string and join by commas
+    # ex: any,contains,I,am,a,search,with,lots,of,commas -> I am a search with lots of commas
```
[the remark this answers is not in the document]
JPrevost: I'll take a look and correct the example as appropriate. This is likely because I added a second commit to change where the collapse of a multi-comma'd search gets handled and didn't update the comment. Good catch!

app/models/preprocessor_primo.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  # extract keyword work at the level of a single keyword query input coming from primo and
+  # returns a string with just that keyword with the operators removed
+  # @param [String] query_part
+  # @return [String] the extracted keyword phrase
+  def self.extract_keyword(query_part)
+    query_part_array = query_part.split(',')
+
+    return 'invalid primo query' unless query_part_array.count >= 3
```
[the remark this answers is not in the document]
JPrevost: Yes. And similarly to the feedback I provided to you on your open PR, maybe this means we log an exception to Sentry as we don't expect this to happen... it's literally exceptional (unless it isn't because it happens to much at which point we don't understand the system and need to fix it)

app/models/preprocessor_primo.rb
JPrevost: I'll take a look. There was a refactor involving this that may have been a bit lazy.

test/models/preprocessor_primo_test.rb
JPrevost: We're composing it in one place that way...but if the method is reused elsewhere or there is a refactor that changes how we are calling it,  it's important to have tests about our expectations for the method independently (unless it's a private method) to ensure we don't break something by changing this behavior.

I'll take a look at this to see if there is a way to make it seem less weird though. I did a sort of late refactor and it's possibly I only half way completed what was in my brain and thus caused this (if you look at the first commit alone and the changes in the second you can probably see how it may have made sense to do it this way at first but might not after the change... I'll check).

test/controllers/graphql_controller_test.rb
```diff
@@ -214,4 +214,30 @@ class GraphqlControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Transactional', json['data']['lookupTerm']['categories'].first['name']
     assert_in_delta 0.95, json['data']['lookupTerm']['categories'].first['confidence']
   end
+
+  test 'primo searches use the preprocessor to extract actual keywords' do
```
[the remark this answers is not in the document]
JPrevost: This feels like a ticket that would sit in a backlog forever and then someone might delete it in a few years after realizing we either already fixed it along the way or it wasn't worth it. I'm not saying you can't open a ticket, I just wouldn't encourage it. This feels like something that when someone is working on this suite and feels like cleaning it up they will more than someone would work on it just because we have a backlog ticket. I'd be happy to chat in a team meeting about whether this type of work should get backlog tickets though as I'm open to different perspectives.

app/models/preprocessor_primo.rb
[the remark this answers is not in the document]
JPrevost: I'll update this to have a Sentry log before merging.

app/models/preprocessor_primo.rb
JPrevost: I think I updated the docs and names of methods to make this more clear what is going on. If not, please let me know.

app/models/preprocessor_primo.rb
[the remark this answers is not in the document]
JPrevost: I've update to send event to Sentry if this unexpected condition happens so we can better understand it

JPrevost:
@matt-bernhardt I've updated the documentation and made some other small changes. Please take a look when you can to see if this fully addresses your concerns.

app/models/preprocessor_primo.rb
```diff
@@ -0,0 +1,76 @@
...
+class PreprocessorPrimo
+  # to_tacos processes raw incoming query from Primo, looks at each part to see if it is a keyword anywhere search
+  # Any portion that is not a keyword anywhere search drops the entire search from TACOS, logging
+  # as the shared Term `unhandled complex primo query` to allow us to track how frequently we are
+  # dropping terms so we can come back later to build out more complex handing if this is common enough
+  # to warrant the additional work.
+  # @param query [String] example `any,contains,this is a keyword search`
+  def self.to_tacos(query)
+    # split on agreed upon joiner `;;`
+    split_query = query.split(';;')
```
matt-bernhardt, line 13: Given the filtering that happens before this, I don't think anything needs an adjustment - but I just peeked at our data and there are about a dozen examples of terms already that have `;;` in them. All of these are obviously from not-Primo, and that traffic would not get routed to this model, but it might be a pathway to keep in mind as we investigate anything marked as "unhandled complex primo query"

app/models/preprocessor_primo.rb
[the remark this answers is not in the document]
matt-bernhardt: Yeah, consistency would be good - but lack of it here isn't a reason for anything to change yet. Thanks for confirming that these won't cause an issue for us in prod.

app/models/preprocessor_primo.rb
matt-bernhardt: Thanks - I think the names now work a little better. I've got other thoughts about the test, but will share them below.

test/controllers/graphql_controller_test.rb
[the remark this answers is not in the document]
matt-bernhardt: Works for me - if it starts bugging me more, I can always set up a quick re-organization of the test file separate from other work, at which point a ticket would be relevant.

test/models/preprocessor_primo_test.rb
matt-bernhardt: Good point, and that's a fair reason to have a test like this.

matt-bernhardt (approved):
Thanks for talking through my questions, for adding a Sentry call on the invalid query branch, and updating the code comments / method names. I think all my concerns have been addressed by this point.

Looking forward to seeing this in production, and continuing to iterate on how to receive traffic from multiple sources. This puts us on a good foundation for that sort of work.

app/models/preprocessor_primo.rb
JPrevost: Hmmm. I'll take a closer look and see if there is a better separator to choose that is less likely to exist in an incoming search (I chose this, it's not something Primo provided by default so we can adjust it). Thanks for noticing that.

JPrevost: `;;;` seems fine. I've asked AdamS to update the primo integration to use that and will push that change here as well.

PR #144, opened 2024-11-20
Finish implementing confirmation workflow

matt-bernhardt:
This finishes implementing the confirmation workflow. Full details are in the commit message, including a somewhat lengthy discussion of some side effects and concerns.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-101

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

added technical debt.

### Documentation

### Testing


JPrevost:
@matt-bernhardt the i18n check is interesting. I'll start a thread on Slack to discuss approaches. TLDR though: it's not a default enabled cop, but our config auto enables new cops even when they are `pending`. I think I like that option as it will alert us when a new cop detects something in our codebase, but when it does I think we should discuss whether we want to adopt it rather than just accepting it. This discussion is what I'll move to slack (i.e. do we want to enforce https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsi18nlocaletexts in this repo).

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+                                      user: current_user
+                                    })
+    set_category(confirmation, params[:confirmation][:category])
+    feedback_for(confirmation.save)
+    redirect_to terms_unconfirmed_path
+  end
+
+  private
+
+  # This method takes the result of the confirmation.save directive above and sets an appropriate flash message.
```
JPrevost, line 27: Minor. Rather than `This method takes the...` I find it easier to read the documentation when it starts with the method name, so my suggestion would be something like `feedback_for takes the...`

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  #
+  # The final else clause is likely to be difficult to provoke, but I have trouble building an if conditional without
```
JPrevost, line 29: It's probably best to avoid first person when writing documentation.

See my comment about using Sentry. I think this documentation line could be removed if we log to Sentry when that state occurs as it should be clear to developers that it's not normal if Sentry is involved.

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  # either a default value or an else clause.
+  def feedback_for(result)
+    if result == true && params[:confirmation][:category] == 'flag'
+      flash[:success] = 'Term flagged for review'
+    elsif result == true
+      flash[:success] = "Term confirmed as #{Category.find_by(id: params[:confirmation][:category]).name}"
+    else
+      flash[:error] = 'Unable to finish confirming this term. Please try again, or try a different term.'
```
JPrevost, line 37: If this is an unexpected state to be in it feels worth throwing an exception to Sentry so we are alerted to it. If this is frequent but we rescue like this and nobody tells us but we thought it could never happen, that could be difficult to detect without a mechanism to inform us.

Something like this should work:

`Sentry.capture_message 'Unexpected error confirming term: #{some maybe useful id of the problem term?}'`

More info in the docs:
https://docs.sentry.io/platforms/ruby/guides/rails/usage/#capturing-messages

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+  end
+
+  # This method catches the RecordNotUnique error in case a duplicate confirmation is ever submitted, and shows an
```
JPrevost, line 41: Are we sure we don't want a user to be able to "fix" a previous confirmation? I'm not sure if this would prevent that (maybe not?).

The use case I have in mind is to in the future allow for someone to click on a category and then go "oops I clicked wrong" and be able to update their confirmation (or delete it). I am fairly certain this won't prevent that, but want to check with you to make sure we both think this is just preventing _multiple_ confirmations but not preventing us from adding the ability to review recent confirmations and changing/removing them (in the future).

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  # appropriate error message to the user.
+  def error_not_unique(_exception)
+    flash[:error] = 'Duplicate confirmations are not supported'
+    redirect_to terms_unconfirmed_path
+  end
+
+  # The confirmation form lists options for each Category record, and then one extra option to flag the term for
+  # removal. "Flag" is not a category, but a separate boolean field on the Confirmation model.
```
JPrevost, line 49: Did you consider having `Flagged for removal` be a category or a boolean on a `Term` itself? (I know that landed in a previous commit and I'm not asking for a change, I'm just curious as something being flagged will be best to be removed from all future queries immediately (to be reviewed by us probably) so putting the flag on Term would make it easy to suppress it from all user facing views via a scope. It could be done also with the flag here, but that requires additional more complex queries any time a Term is being looked up. Not a big deal either way, I'm mostly just curious what was considered when making this decision.

I'm nervous to actually `delete` flagged queries every because by keeping them but suppressing them allows us to always hide them from users whereas deleting them would just allow them to resurface again if someone were ever to search for that term again. I guess our review of suppressed queries might have some thought into how likely it would be to recur and need to be re-suppressed again. Hmmm.

JPrevost (changes requested):

app/controllers/confirmation_controller.rb
matt-bernhardt: I was tempted to send something to Sentry, both for this message as well as the RecordNotUnique catch, but then noticed that we don't have it set up. I wasn't sure if that should be a side effect to this work, or if it should be something that gets specific focus?

app/controllers/confirmation_controller.rb
matt-bernhardt: Editing an existing record in this way shouldn't trip the uniqueness requirement, AFAICT - the constraint is on the pair of Term and User, and any edit would only affect the category value.

I agree that this use case would be useful, and also that it is future work. For now, I don't think there's anything here that would preclude that happening.

app/controllers/confirmation_controller.rb
JPrevost: Ah... we do have a ticket for that. I forgot. Hmmm. I'd be okay with opening a ticket to add the Sentry logging later (or puling that Sentry ticket and putting in in place before merging this).

app/controllers/confirmation_controller.rb
matt-bernhardt: I remember discussing a few options, but I'm not sure I could find where now. I do remember thinking about whether we needed 3, 4, or 5 categories - and settling on 4 with the addition of `Undefined`, which humans can use but the knowledge graph does not. A `Flagged` category didn't seem to make sense, based on the assumption that we would actually be deleting terms. If we don't actually delete terms, then perhaps that might make sense to revisit.

My thought about suppressing versus deleting terms is that we shouldn't store records we shouldn't have - keeping them in the database invites problem if the application leaks data or is breached. If a term gets re-submitted, we'll need to delete it again, but I'm not sure it's worth keeping things around in case a duplicate ever gets submitted (remembering how rare repeated searches are in the first place).

I do remember thinking about a flag on the Term, and decided against that approach because having the flag on the Confirmation preserves information about who flagged it - in case follow-up is needed about why they flagged it.

app/controllers/confirmation_controller.rb
matt-bernhardt: Given that we both thought about leveraging Sentry here, I think it's appropriate to pull TCO-88 now, get that done, and then update this.

app/controllers/confirmation_controller.rb
JPrevost: Yeah this is complicated. There aren't many reasons we'd want to flag/delete a term and if we hit one I'm nervous to not have a way to prevent it from being shown to all future validators, keep it out of any data exports, etc. I understand your concern about wanting to delete data we shouldn't have, but I'm waffly on whether it's better to know we shouldn't have it (by flagging it as suppressed at the Term level) to make sure it's never included in our work versus having it continue to show up again in the future. I can probably be brought around to your view based on the infrequency of repeat terms, but by deleting things -- contrary to how it may feel -- we'll never actually know for certain we are preventing it from being seen from anyone (other than us as system maintainers who have raw db access). Hmm.

app/controllers/confirmation_controller.rb
matt-bernhardt: Now that #146 has merged, I'm going to rebase this branch and add the Sentry message here.

matt-bernhardt:
Okay, @JPrevost - I've pushed commits that resolve many of your concerns, and extends the (already-existing, it turns out) exclusion of the i18n cop to include the Confirmation controller. That resolves the issue for CodeClimate.

Remaining is the question of how to handle flagged Terms. My thoughts on that, at the moment, are these:

* Nothing in this PR actually implements what we do about flagged terms - so I don't know that we need to decide in this PR whether we're going to delete flagged terms, or build a way to suppress them. Either option will require development work beyond the current scope of this ticket.

* Given the current data model for Confirmations, I think the PR as currently structured is the way to implement the rest of the confirmation workflow. The question is whether we want to change the data model or not, which leads me to:

* I'm ambivalent about whether to switch to five categories, or to stay with four categories and a boolean. Having five categories would make this PR simpler, which is somewhat compelling. We'd drop the boolean flag from the Confirmation model, and add a fifth seed value - updating tests, controller, and view templates along the way. I do have a pretty strong preference that the updated data model continue to store who flagged the term - which I think we'd get with a fifth category.

What are your thoughts at this point?

JPrevost:
@matt-bernhardt this may be something best discussed synchronously, but I'll try to share my thoughts here.

I think the best place to flag a Term we are concerned about is in the Term model itself.

I don't think using a Category of "Flagged for review" or a boolean on a confirmation provides the affordances we'll want to suppress the Term from being used until we have reviewed and possibly chosen to delete it.

A future scope on Term "without_flagged" (other names might be "displayable" or something that makes it clear what the intent is) could be used consistently throughout the application to prevent anyone from seeing flagged results until such time as we as data maintainers either remove the flag as erroneous or delete the Term.

Using a Category or boolean flag on a Confirmation could also be used to create a scope, but it feels much more complicated to reason about as any time we use a Term the scope would need to look to other tables to determine if it is displayable. As we'll have so few flagged Terms, this seems like it's more work with minimal gain.

With all that said:

- I agree with your statement that actually deleting (or not) is out of scope of this work. I still think that means we should assume we'll want to either suppress or delete later and make decisions that feel most conducive to that as if we don't plan on doing anything ever then we might as well not flag at all.
- I would be okay with removing the ability to flag from this PR entirely and opening a ticket to plan that feature more thoroughly (including deciding if we will immediately suppress and delete later or something else for flagged Terms)
- I would be okay with creating a boolean on Term in this PR
- I would be okay with using a Category to flag Terms
- I would accept, but anticipate changing in the future, using a boolean on Confirmation. My reluctant acceptance comes down to it being fairly easy to change it later. It still feels like the incorrect data model to me as we aren't flagging a Confirmation, we are flagging a Term... but it _will_ work and when we build out the feature more thoroughly we can change it.

matt-bernhardt:
I can see a solid rationale for adopting a scope on the Term model for "records that have not been flagged" - and that we would soon want to enforce that scope in nearly every interaction where the application asks for a list of Terms. This also feels like it calls for an interface to show which records have been hit by this flag, although the interface doesn't feel like it's in scope of the current ticket.

I'm less clear whether the _creation_ of this scope would be part of this ticket, although the conversion of all existing references to use the scope feels like it would be future work.

What about something like this as a middle ground, at least for the short term but possibly for longer:

1. Change the data model here to create a Flagged category, and update the Confirmation model to refer only to the category value (dropping the flagged boolean field).
2. Create a new `flagged` field on the Term data model, with the future intent to hang the scope and everything else off _this_ field.
3. Update the `create` method in this PR to set the `Term.flagged` field in addition to recording the details of the user's Confirmation submission.

This feels like it would achieve both of our goals:
* The user's Confirmation would be preserved, allowing us to follow up with needed about why a flag was thrown. Having the field as part of the user submission also allows us to support Confirmations from multiple users without loss of data.
* Simultaneously setting a flagged boolean on the Term model serves the operational needs of the application in a more efficient way, as you note that leveraging this scope will be much simpler in database terms.

Both the record of the user's activity, and the operation needs of the application, are preserved - at the cost of a slight duplication of information. However, this duplication will come in handy if we ever decide that the flag is not justified, because we can release a term for processing while still having some breadcrumb that it was once flagged (just in case that's ever a question).

I'd be comfortable with this as a route forward, and am willing to refactor the current PR toward this goal. There would then be future tickets to build the scope, leverage it across the application, and build an interface to show admins (only) what terms have been flagged.

JPrevost:
@matt-bernhardt Yup. That makes a ton of sense to have both the Category and Term boolean for the reasons you stated. Thanks for talking this through to help figure out what made sense to build towards future features versus what the future features are that we can ticket but not implement at this time.

matt-bernhardt:
I'm thinking that rebasing this branch past the recent merges will be fine, as long as I keep the current three commits unchanged - but I want to confirm this before force-pushing.

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  private
+
+  def confirmation_params
+    params.require(:confirmation).permit(:term_id, :category_id, :user_id)
+  end
+
+  # feedback_for takes the result of the confirmation.save directive above and sets an appropriate flash message.
+  #
+  # The final else clause is likely to be difficult to provoke, so we are sending a Sentry message in that block in
```
JPrevost, line 38: I think this line of documentation should be removed or moved as it does not reflect the current state of the method it is documenting.

JPrevost (approved):
Looks good. Two minor suggested documentation updates (one to remove/move a line that is no longer accurate should definitely be fixed, the other is a suggestion.

app/controllers/confirmation_controller.rb
matt-bernhardt: Good catch - I've rewritten this entire comment block. Thanks!

app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+  def new
+    @term = Term.find(params[:term_id])
+    @categories = Category.all
+    @confirmation = Confirmation.new
+  end
+
+  # create receives the submission from the new confirmation form, creating the needed record with the help of various
+  # private methods.
+  def create
+    new_record = Confirmation.new(confirmation_params)
```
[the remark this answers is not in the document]
matt-bernhardt: When I saw what the final diff ended up being, I agree that it makes things much simpler. I'm not completely happy that it doesn't seem to be possible to chain `.require()` statements in the way I originally expected, but this ended up being a bit better than I at first feared.

PR #157, opened 2024-12-12
Add devcontainer config

JPrevost:

* Providing a standard development environment will allow for easier
  onboarding and diagnosis of whether an issue is a single local dev
  problem or also a problem in our standard environment

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-89

* Adapts the devcontainer approach that new Rails apps provide for TACOS
* Provides some light documentation in the README about alternatives for
  developers that are not using VScode

Removes an unused Dockerfile

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

- [x] I have confirmed that the code works as intended.
added technical debt.

### Documentation

### Testing

- [x] No additional test coverage is required.


matt-bernhardt (approved):
With the caveat that I'm not super fluent with Docker compose files, I can confirm that the devcontainer defined here works when I'm using VSCode, and I've been able to start it up, run tests and start the server, etc.

The documentation looks good, and addresses my initial confusion about the container being distinct from the server process inside the container.

Given all the above, I'm comfortable approving this - although if you'd rather get an approval from someone more familiar with containers, I understand.

My only remaining question, which is probably addressed by future-us rather than now-us, is whether any of us are intending to work this way at the moment? I'm slightly concerned about a future state where we've added this container for easier onboarding, but it has broken at some point and no one notices because none of us are dogfooding it.

JPrevost:
@matt-bernhardt Good point re: whether we'll actually use this and nod to maintainabiltiy questions. For context, this is all default for upcoming versions of Ruby on Rails and I just yanked it from a more recent version and popped it into ours. It largely points to external config that is maintained by the community so it _should_ sort of maintain itself. 🤞🏻

PR #162, opened 2024-12-19
Split Detector::Journals model into two, extending BulkChecker into the detection model

matt-bernhardt:
This splits the `Detector::Journal` model into two pieces:

1. `Journal` - which is concerned with storing the list of journal records which incoming terms are compared with. This is a database-backed Rails model, outside the detector namespace.
2. `Detector::Journal` - which actually does the comparison between incoming terms and the list of records. This is a plain ruby object, no backing database. It now includes the BulkChecker module, so developers can use the `.check_all_matches` method for their work.

In the process of making these changes, the existing set of journals which were imported via a rake task are dropped, so will need to be re-imported. The rake task has been updated to deposit this information in the new `Journal` model, but that population _is not_ part of this PR. I left it out from a concern that the migration would take too long during deployment. The impact of this is that, for the minutes between this change being deployed and when the new rake task is run, TACOS will not match any journals.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-110

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

YES migrations are included 

**PLEASE NOTE** - running these migrations will result in the loss of data from your local application, as any existing `Detector::Journal` records will be lost. Rolling back the migration is possible, but not the restoring of lost records.

### Code

added technical debt.

### Documentation

### Testing

- [x] No additional test coverage is required.


jazairi (approved):
I haven't confirmed the BulkChecker functionality, but this approach aligns with my understanding of what we're trying to do here. Having a model for how to separate out the ActiveRecord bit will make this easier to implement for suggested resources, so thanks for kicking this off!

PR #171, opened 2025-01-14
Validate ISBNs detected via regex

JPrevost:

* It is not possible to validate detected ISBNs via regex alone
* We are seeing false positives for ISBNs in production that are invalid
* We validate ISSNs, but had not gone back and added this logic to ISBN

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TCO-114

* Follows the pattern to strip detected ISBNs in a way similar to ISSNs
* Follows the ISBN-10 and ISBN-13 validation specifications
* Updates tests for valid ISBNs are actually using valid ISBNs (previously, the ISBN-10s were incorrect converted to ISBN-13s)
* Adds additional ISBN-13 examples to test positive validations
* Adds additional ISBN-13 examples to test negative validations

* Removed "ISBN: " optional portion of the ISBN detection. It didn't feel useful and needed to be stripped out before validation (which felt like it supported to me that it wasn't part of the ISBN and shouldn't be detected)
* Updated yardopts to include private methods. We do a lot of important work in private methods. While we shouldn't call them from outside of the instance, having them in the docs feels better as some of our docs were effectively making Classes look like black boxes without this change.
* Some of our test ISBN-13s were invalid. The ISBN-10 list was run through an external ISBN-10 to ISBN-13 convertor. As all of those will start with 978 prefixes, I manually added some valid ISBN-13 examples to the tests that started with the other valid prefix (979).

Example query. Run in prod, this will show a detected ISBN that then fails the details lookup as it isn't actually valid. In the PR build or locally with this branch, it does not detect the invalid ISBN

```graphql
{
  logSearchEvent(
    searchTerm: "Artin, M. Algebra (2nd Edition). Addison Wesley, 2010. ISBN: 9780132413771"
    sourceSystem: "playground"
  ) {
    phrase
    categories {
      name
      confidence
    }
    detectors {
      standardIdentifiers {
        kind
        value
        details {
          linkResolverUrl
        }
      }
    }
  }
}
```

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-###

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/detector/standard_identifiers.rb
+++ b/app/models/detector/standard_identifiers.rb
@@ -17,12 +17,15 @@ class Detector
     extend Detector::BulkChecker
 
     # Initialization process will run pattern checkers and strip invalid ISSN detections.
-    #   @param phrase String. Often a `Term.phrase`.
-    #   @return Nothing intentional. Data is written to Hash `@detections` during processing.
+    #
+    # @param phrase String. Often a `Term.phrase`.
+    # @return nil. Data is written to Hash `@detections` during processing. Things technically get
+    #   returned here but it is a side effect and should not be relied on.
     def initialize(phrase)
       @detections = {}
       pattern_checker(phrase)
       strip_invalid_issns
+      strip_invalid_isbns
     end
 
     # The record method will consult the set of regex-based detectors that are defined in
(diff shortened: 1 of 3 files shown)


app/models/detector/standard_identifiers.rb
```diff
@@ -53,13 +56,76 @@ def self.record(term)
...
+        false
+      end
+    end
+
+    # valid_isbn_10? follows the ISBN 10 specification for validation
+    # https://en.wikipedia.org/wiki/ISBN#ISBN-10_check_digits
+    #
+    # @param digits Array. An array of strings representing each character from a detected ISBN candidate.
+    # @return boolean
+    def valid_isbn_10?(digits)
```
JPrevost, line 102: There's an alternate way to write this taking advantage of the reduce method. I wasn't sure which felt more clear and chose not to use reduce but wanted to show the other option to get input from code review as to which feels easier to understand (as they both do the same thing).

`acc` is a conventional name for the accumulator. If another name would be clearer, there is nothing special about `acc`.

```ruby
def valid_isbn_10?(digits)
  sum = digits.each_with_index.reduce(0) do |acc, (digit, index)|
    digit = '10' if digit.casecmp('x').zero?
    acc += digit.to_i * (10 - index)
  end
  (sum % 11).zero?
end
```

app/models/detector/standard_identifiers.rb
```diff
@@ -53,13 +56,76 @@ def self.record(term)
...
+      end
+      (sum % 11).zero?
+    end
+
+    # valid_isbn_13? follows the ISBN 13 specification for validation
+    # https://en.wikipedia.org/wiki/ISBN#ISBN-13_check_digit_calculation
+    #
+    # @param digits Array. An array of strings representing each character from a detected ISBN candidate.
+    # @return boolean
+    def valid_isbn_13?(digits)
```
JPrevost, line 116: Similar to the comment above, this is an example using reduce for input on whether one feels easier to understand on a quick read of the method.

```ruby
def valid_isbn_13?(digits)
  sum = digits.map(&:to_i).each_with_index.reduce(0) do |acc, (digit, index)|
    acc += digit * (index.even? ? 1 : 3)
  end

  (sum % 10).zero?
end
```

matt-bernhardt (commented):
I have a few thoughts about this, nothing significantly blocking but I one question I wanted to get your thoughts on prior to approval. The rest is conversational.

The question / request is whether we need a test for the ISBN validation that's similar to the ISSN test at https://github.com/MITLibraries/tacos/blob/main/test/models/detector/standard_identifiers_test.rb#L93-L124 - which feels like a thoroug confirmation that the check digit validation is correct compared to testing a set of (apparently random?) values.

---
Beyond that comment, my thoughts about the variations on the validation logic are that I'm fine with either using the `reduce` method or leaving it out, although my brain latched onto the method without more easily. Having now looked up the `reduce` documentation, it feels similar to a ternary - a simpler syntax that's relatively easy to follow, but something to keep in mind. It's also relatively easy to look up if I find myself forgetting it.

The bit that trips me up the most, actually, is remembering that `digit.casecmp('x').zero?` returns either 0 or -1, so there's a bit of parsing there which I find `digit.downcase == 'x'` avoids entirely. Was that flagged by a linter?

Other thoughts:
* I like the addition of private methods to the yardopts - I agree that there are private methods in our application which feel awkward to not have documented (I think we've been writing some already, but including them for future-us in this way feels right)
* Seeing the amount of validation logic growing in the `StandardIdentifiers` class makes me wonder whether the validation might be more suitable in the lookup classes as an public method? That might allow this class to purely focus on the regexes and pattern matching, while leaving all the "rubber meets the road" logic for each format in its own models. For the moment I think the class is fine, at least.
* Thank you for parsing the math on these validations - the wikipedia articles you've referenced can (as usual for wikipedia) be a bit dense to come to terms with, but I think the appliation is better for it.

JPrevost:
@matt-bernhardt yes, the linter suggested `casecmp`. It's listed as a performance cop so I'm inclined to stick with it even if it is less readable.

I'll not introduce reduce with this work. I think it's such a toss up to just go with what we both feel reads easier.

I'll add an explicit test of ISBNs with invalid check digits rather than implicitly testing that via the couple I added.

Re: refactoring StandardIdentifiers to pull validation out. Yes, that makes sense. I'm not sure exactly how to approach that yet but am in support of it. For a while I wanted to remove it entirely but I think it might make sense to keep it for the regexes as you noted and move the validation to separate classes. Maybe Refactoring the Lookup classes to not just be lookups. So LookupISBN becomes ISBN which can handle validation and lookups rather than just adding to the Lookup classes as-is. It feels if they are changing scope the naming may change too. We'll see. I like the gist of your idea though.

matt-bernhardt (approved):
If there's a performance implication significant enough that the linter flags it, then I'll deal with the complexity of parsing `digit.casecmp('x').zero?` . TIL.

Thanks for adding the negative tests - this all looks good to me!

:shipit:

PR #174, opened 2025-01-22
Add option to confirm only categorized terms

matt-bernhardt:
### Why are these changes being introduced:

The confirmation workflow was predicated on humans providing a second opinion on categorization, allowing us to review how the application is already performing. However, when we initially rolled out this workflow, it allowed humans to provide confirmation on every term in the system, and not just those which the system had categorized.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/tco-126

### How does this address that need:

This adds:

* A categorized scope to the Term model

* Forks the logic in the term controller's "unconfirmed" method, returning two different recordsets. By default it includes only categorized terms (via the scope added above). However, there is a URL parameter to leave this restriction out.

* The template for this list of terms has buttons to switch between the two modes. The default behavior applies the categorized scope, while we can still see all terms via the other button.

* We add tests for both the model scope and the branching logic in the controller, making sure that they behave as expected.

### Document any side effects to this change:

* I'm a little concerned that I've missed writing tests in some cases, but I think this covers most of the feature.

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/controllers/term_controller.rb
+++ b/app/controllers/term_controller.rb
@@ -3,8 +3,15 @@
 class TermController < ApplicationController
   include Pagy::Backend
 
-  # This provides the list of terms that are awaiting confirmation.
+  # This provides the list of terms that are awaiting confirmation. By default this shows only terms which have been
+  # categorized automatically. Adding `type=all` to the querystring will show _all_ terms which the user has not yet
+  # confirmed.
   def unconfirmed
-    @pagy, @records = pagy(Term.user_unconfirmed)
+    terms = if params[:show] == 'all'
+              Term.user_unconfirmed
+            else
+              Term.categorized.user_unconfirmed
+            end
+    @pagy, @records = pagy(terms)
   end
 end
--- a/app/controllers/term_controller.rb
+++ b/app/controllers/term_controller.rb
@@ -8,6 +8,7 @@ class TermController < ApplicationController
   # confirmed.
   def unconfirmed
     terms = if params[:show] == 'all'
+              authorize! :confirm_uncategorized, Term
               Term.user_unconfirmed
             else
               Term.categorized.user_unconfirmed
--- a/app/views/term/unconfirmed.html.erb
+++ b/app/views/term/unconfirmed.html.erb
@@ -3,10 +3,12 @@
 <p>The terms listed here have not yet been reviewed by a person and placed into a category. Clicking on any term will
 take you to the form for submitting this information.</p>
 
+<% if can? :confirm_uncategorized, Term %>
 <p class="wrap-filters">View:
   <a class="btn button-small button-secondary" href="<%= terms_unconfirmed_path %>">Categorized terms</a>
   <a class="btn button-small button-secondary" href="<%= terms_unconfirmed_path(show: 'all') %>">All terms</a>
 </p>
+<% end %>
 
 <p><%== pagy_info(@pagy) %></p>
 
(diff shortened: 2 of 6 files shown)


app/controllers/term_controller.rb
```diff
@@ -3,8 +3,15 @@
 class TermController < ApplicationController
   include Pagy::Backend
 
-  # This provides the list of terms that are awaiting confirmation.
+  # This provides the list of terms that are awaiting confirmation. By default this shows only terms which have been
+  # categorized automatically. Adding `type=all` to the querystring will show _all_ terms which the user has not yet
+  # confirmed.
```
JPrevost, line 8: We don't need to change for this PR as the default is what we expect people to use for the foreseeable future, but it might be better longterm to make this a sticky option by setting it as a session variable or cookie. I don't even think we need a ticket for that as we may never come back to it, but just wanted to share that the ergonomics of this is a bit awkward if someone is trying to "categorize" terms rather than validate categorizations.

app/views/term/unconfirmed.html.erb
```diff
@@ -3,6 +3,11 @@
 <p>The terms listed here have not yet been reviewed by a person and placed into a category. Clicking on any term will
 take you to the form for submitting this information.</p>
 
+<p class="wrap-filters">View:
```
JPrevost, line 6: I'm not sure if this UI option should be displayed to all users. I believe our goal is to just have people validate for the foreseeable future and this may make that slightly more confusing.

I'd recommend either hiding this behind admin capabilities or removing it entirely. Having the feature available I like...presenting the option to all logged in users feels like we are possibly inviting trouble.

test/models/term_test.rb
```diff
@@ -299,6 +299,32 @@ class TermTest < ActiveSupport::TestCase
...
+    new_record = {
+      term: t,
+      category: categories('navigational'),
+      confidence: 0.5,
+      detector_version: '1'
+    }
+    Categorization.create!(new_record)
+
+    # The term has gained a category, but the categorized scope has not changed size.
+    assert_operator term_category_count, :<, t.categorizations.count
```
JPrevost, line 324: No change necessary, but `term_category_count` as a variable name made me have to think more than felt necessary as to what is going on here. Maybe something like `orig_category_count` as it gets compared to the same count after a categorization is made might be better?

JPrevost (changes requested):
I'd like to see the UI affordance of selecting between "verification" mode and "categorization" mode removed from default logged in users. If you feel strongly that everyone should get this affordance, I'd be open to listening as to why and revise my review to approved.

test/models/term_test.rb
matt-bernhardt: Yeah, I can see how a different name might be better. I've tried to make it something clearer, and I particularly like the `orig_` prefix.

app/controllers/term_controller.rb
matt-bernhardt: Yeah, having the choice be sticky seems like something that could be useful in the future, or even a workflow that just picks terms according to some logic, so that the user never goes back to a list-of-terms page.

Regardless, I agree that this is a concern for the future should we ever really start using this workflow.

app/views/term/unconfirmed.html.erb
matt-bernhardt: I've updated the implementation to enact a restriction like this, so basic users don't see (and can't use) the broader scope. I've also written a number of new controller tests to confirm that things behave the way I expect. I'm not particularly happy with having a test that's this tightly coupled to UI text, but I'd rather this than not have a test for it at all.

JPrevost (approved):

PR #145, opened 2024-11-22
Introduce ADR for fingerprinting

jazairi:
### Why these changes are being introduced:

We currently use fingerprinting for suggested resources. We are planning to expand this functionality to terms to support clustering, and we also need to refactor suggested resource fingerprinting to accommodate multiple phrases/fingerprints per suggested resource. This felt like a good opportunity to reevaluate our fingerprinting implementation.

### Relevant ticket(s):

* [TCO-74](https://mitlibraries.atlassian.net/browse/TCO-74)
* [Add automatic fingerprinting for Term records (unticketed -- link to PR)](https://github.com/MITLibraries/tacos/pull/138)

### How this addresses that need:

This ADR proposes that we add a central `Fingerprint` model, linked to `Terms`, which is leveraged by `SuggestedResource` and any other detectors that require fingerprinting.

EngX had discussed this as a possible approach in a recent meeting. Consensus on this ADR would confirm the decision.

### Side effects of this change:

See ADR for details.

### Ticket(s)

See above.

### Accessibility

- [ ] ANDI or Wave has been run in accordance to [our guide](https://mitlibraries.github.io/guides/basics/a11y.html) and
      all issues introduced by these changes have been resolved or opened 
      as new issues (link to those issues in the Pull Request details above)
- [x] There are no accessibility implications to this change

### Documentation

- [ ] Project documentation has been updated, and yard output previewed
- [ ] No documentation changes are needed

### ENV

- [ ] All new ENV is documented in README.
- [ ] All new ENV has been added to Heroku Pipeline, Staging and Prod.
- [z] ENV has not changed.

### Stakeholders

- [ ] Stakeholder approval has been confirmed
- [x] Stakeholder approval is not needed

### Dependencies and migrations

NO dependencies are updated

NO migrations are included

### Code

- [ ] I have confirmed that the code works as intended.
- [ ] Any CodeClimate issues have been fixed or confirmed as
added technical debt.

### Documentation

- [ ] The commit message is clear and follows our guidelines
      (not just this pull request message).
- [ ] The documentation has been updated or is unnecessary.
- [ ] New dependencies are appropriate or there were no changes.

### Testing

- [ ] There are appropriate tests covering any new functionality.
- [ ] No additional test coverage is required.

[...]

PR #175, opened 2025-01-22
Rails 7.2 upgrade

JPrevost:
Summary of changes (please refer to commit messages for full details)

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-###

### Accessibility

- [ ] ANDI or Wave has been run in accordance to [our guide](https://mitlibraries.github.io/guides/basics/a11y.html) and
      all issues introduced by these changes have been resolved or opened 
      as new issues (link to those issues in the Pull Request details above)
- [x] There are no accessibility implications to this change

### Documentation

- [ ] Project documentation has been updated, and yard output previewed
- [x] No documentation changes are needed

### ENV

- [ ] All new ENV is documented in README.
- [ ] All new ENV has been added to Heroku Pipeline, Staging and Prod.
- [x] ENV has not changed.

### Stakeholders

- [ ] Stakeholder approval has been confirmed
- [x] Stakeholder approval is not needed

### Dependencies and migrations

YES dependencies are updated

YES migrations are included

### Code

- [ ] I have confirmed that the code works as intended.
- [ ] Any CodeClimate issues have been fixed or confirmed as
added technical debt.

### Documentation

- [ ] The commit message is clear and follows our guidelines
      (not just this pull request message).
- [ ] The documentation has been updated or is unnecessary.
- [ ] New dependencies are appropriate or there were no changes.

### Testing

- [ ] There are appropriate tests covering any new functionality.
- [ ] No additional test coverage is required.

PR #185, opened 2025-02-10
Refactor SuggestedResource to leverage new Fingerprint model

jazairi:

We [recently decided](#145) to make a separate Fingerprint model, associated with Term, as multiple detectors are likely to use fringerprinting (implemented in #138). We have also begun to split the ActiveRecord components of detectors into separate models (implemented for Detector::Journal in #162).

### Relevant ticket(s):

* [TCO-111](https://mitlibraries.atlassian.net/browse/TCO-111)
* [TCO-122](https://mitlibraries.atlassian.net/browse/TCO-122)

* Splits the ActiveRecord components of Detector::SuggestedResource into a separate SuggestedResource model.
* Associates SuggestedResource with Fingerprint, via Term, such that a suggested resource can have multiple terms and fingerprints.
* Removes the suggested resource dashboard (see side effects).

* Terms that are associated with a suggested resource should not be destroyed. Rails does not allow the `:dependent` option on `belongs_to` associations, so this commit instead adds a `before_destroy` hook with a custom method that aborts the callback and logs the attempt in Sentry.
* Because administrate does not handle has_many relationships well, we will need to build a custom dashboard to manage suggested resources. This is ticketed as [TCO-145](https://mitlibraries.atlassian.net/browse/TCO-145). Until that UI is ready, we will use the Rails console to make any requested changes to suggested resources.

### Accessibility

### Documentation

### ENV

- [x] ENV has not changed.

### Stakeholders

- [x] Stakeholder approval has been confirmed
- [ ] Stakeholder approval is not needed

NO dependencies are updated

YES migrations are included

### Code

added technical debt.

### Documentation

### Testing

[TCO-111]: https://mitlibraries.atlassian.net/browse/TCO-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TCO-122]: https://mitlibraries.atlassian.net/browse/TCO-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TCO-145]: https://mitlibraries.atlassian.net/browse/TCO-145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/metrics/algorithms.rb
```diff
@@ -17,7 +17,6 @@
 #  suggested_resource_exact :integer
 #  lcsh                     :integer
 #  citation                 :integer
-#  barcode                  :integer
```
JPrevost, line 20: I think this is an incorrect outcome of a rebase at some point?

jazairi: It's definitely wrong, because the migration adding `barcode` still exists in the current commit, and I see the field in `schema.db`. But, running `annotate --models` again doesn't reinsert it. I'll add it in manually, but I'm a little disconcerted by this.

app/models/suggested_resource.rb
```diff
@@ -0,0 +1,43 @@
...
+  has_many :fingerprints, through: :terms, dependent: :nullify
+
+  # This replaces all current SuggestedResource records with a new set from an imported CSV.
+  #
+  # @note This method is called by the suggested_resource:reload rake task.
+  #
+  # @param input [CSV::Table] An imported CSV file containing all Suggested Resource records. The CSV file must have
+  #                           at least three headers, named "Title", "URL", and "Phrase". Please note: these values
+  #                           are case sensitive.
+  def self.bulk_replace(input)
```
JPrevost, line 27: @jazairi 

In order to proceed with the review more efficiently, I came up with what _might_ be okay for this `bulk_replace` method.

```ruby
  def self.bulk_replace(input)
    raise ArgumentError.new, 'Tabular CSV is required' unless input.instance_of?(CSV::Table)

    # Need to check what columns exist in input
    required_headers = %w[title url phrase]
    missing_headers = required_headers - input.headers
    raise ArgumentError.new, "Some CSV columns missing: #{missing_headers}" unless missing_headers.empty?

    SuggestedResource.destroy_all

    input.each do |line|
      term = Term.find_or_create_by(phrase: line['phrase'])
      term.suggested_resource = SuggestedResource.new({ title: line['title'], url: line['url'] })
      term.save
    end
  end
```

It changes the required_headers to lowercase as that's how they come out of Heroku DataClips (there is a saved DataClip with the prod data) and avoids the problem I ran into by starting with Term and adding SuggestedResources instead of the other way around which was a remnant from before they were related.

JPrevost: I did not adjust tests that are likely failing now, nor do a ton of manual verification they actually migrated to the expected terms, but I now have 117 SuggestedResources locally I can use to help understand these changes better which was my main goal :)

JPrevost: Ah. This approach ends up with a 1:1 mapping of `Terms` to `SuggestedResources` rather than linking `SuggestedResources` to multiple `Terms` which was half the battle.

jazairi: Here's what I was tentatively working with, but I haven't confirmed it yet:

```ruby
    input.each do |line|
      record = SuggestedResource.find_or_create_by({ title: line['Title'], url: line['URL'] })
      record.save
      record.terms.find_or_create_by(phrase: line['Phrase'])
    end
```

But again, I haven't tested that yet, and it seems too easy to work. In that case, I'm fine working from the term instead. If you want to push what you have, I can write/edit the tests. Thanks for looking into this!

JPrevost: I've pushed what I had, if after you look more closely at it your version does the same thing it's definitely way simpler and we should go with it instead :)

app/models/term.rb
```diff
@@ -104,6 +107,16 @@ def check_fingerprint_count
...
   end
 
+  # This is called before_destroy to avoid orphaning SuggestedResource records. Deleting terms should be an unlikely
+  # event, so this should come up rarely. If it does, it warrants the extra care to delete the record manually in the
+  # Rails console.
+  def check_suggested_resource
+    return unless suggested_resource
+
+    Sentry.capture_message('Cannot delete term with associated suggested resource')
+    throw :abort
```
JPrevost, line 117: I'd suggest also adding something like `Rails.logger.error('Cannot delete term with associated suggested resource')` to allow more immediate feedback when run in a console (which seems like our likely current use).

development.rb seems to be missing a logging line that makes this actually work well since we upgrade to 7.2. But if I add [this back in](https://github.com/MITLibraries/tacos/pull/175/files#diff-d3c4b3f41072daa416f1920511e9b2e26caea8c5cec0a14cb9508589a4dafa47L77-L79) logging does seem to work.

test/models/term_test.rb
```diff
@@ -448,4 +449,18 @@ class TermTest < ActiveSupport::TestCase
...
+    assert term.suggested_resource
+
+    term.destroy
+    assert term.present?
+  end
+
+  test 'destroys record successfully if no suggested resource is present' do
+    term = terms(:hi)
+    assert_nil term.suggested_resource
+    assert_nothing_raised { term.destroy }
```
JPrevost, line 464: Nothing is raised in the other test so Im not sure this is testing what we think it is? Maybe a better assertion is to confirm `Term.where(phrase: 'whatever the original term we just deleted phrase was')` returns no match?

JPrevost: Or maybe just follow what the other test does and check for `term.present?` If that doesn't work as expected then both tests would need updating :)

JPrevost (commented):
A few small tweaks suggested, but this seems to work exactly like we wanted. Thanks for working on this!

db/migrate/20250107204800_create_suggested_resources.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSuggestedResources < ActiveRecord::Migration[7.1]
```
JPrevost, line 1: @jazairi I'm not sure how important this is, but the order of migrations is currently incorrect. I noticed this when trying to rollback the latest migration to get back to `main` and it rolled back 2 migrations that landed last week before getting to these 3.

I suspect it would be better to have these named such that they show up after the latest `main` migrations as I believe Rails just tracks the latest date migrations have completed and not each individual migration status so there is risk that some environments will skip these migrations and we'll be super confused.

I may be misunderstanding how this works though...

test/models/term_test.rb
jazairi: Oh yeah, good catch. I think this was from an earlier iteration where I was raising an exception rather than aborting the callback. I agree that `assert_not term.present?` is much better here.

db/migrate/20250107204800_create_suggested_resources.rb
jazairi: Gotcha. Is that as simple as changing the date marker to, say, today?

JPrevost: I think so 🫰

JPrevost: (GitHub has the worst fingers crossed emoji. That's what that is supposed to be)

jazairi:
@JPrevost I think the latest commit covers everything, but please let me know if I've missed something. Specifically, I'm a little shaky on whether the new test coverage is adequate/accurate.

jazairi:
I've never seen this GitHub actions caching error. Not sure what that's about. I'm going to try closing/reopening the PR to kickstart CI again... 🤞

test/tasks/suggested_resource_rake_test.rb
```diff
@@ -52,19 +52,63 @@ def setup
...
+  end
+
+  test 'reload handles phrases that match existing terms' do
+    records_before = Term.count
+    VCR.use_cassette('suggested_resource:reload from remote csv') do
+      remote_file = 'http://static.lndo.site/suggested_resources.csv'
+      Rake::Task['suggested_resources:reload'].invoke(remote_file)
+    end
+
+    # The CSV has two rows, one of which is an existing term (web of science). So, only one term should be created.
```
JPrevost, line 111: Non blocking. You asked about tests so I spent a bit more time than I normally might with these.

The only suggestion I have is that we might confirm for each of these the change in Term count _and_ the change in SuggestedResource count both match the expectations as sometimes only once changes, sometimes both.

The other comment would be that I might test the bulk_loader method more directly rather than the rake task which just adds complexity. This looks like a carry-forward from the original implementation and just updated to the new spec so I see why you chose this option for this work as changing behavior _and_ how we test is more risky so for now this is probably best to keep as-is (and maybe add assertions for SuggestedResource count if it feel it adds value to track both.

JPrevost (approved):
Once we've confirmed with UX the plan for us to manually make changes is okay, we're good to merge on my end.

test/tasks/suggested_resource_rake_test.rb
jazairi: This is a good point, and as I was looking back into this, I realized that even the assertions we _have_ might not be matching the expectations. Sigh. I want to make sure we get this right before merge, so I'm going to double-check all of these. I'll add the recommended assertions back in as part of that review.

jazairi: @JPrevost I updated the tests to include assertions on both record types, and I updated the file fixtures (and cassettes) to more accurately reflect the cases we're trying to test. I know you've already approved, but I'm going to re-request review because this feels like a significant enough change to warrant it.

JPrevost: Great change, thanks!

JPrevost (approved):

PR #243, opened 2025-06-02
Integrate with external citation detector via Detector::MlCitation class

matt-bernhardt:
Now that the external citation detector is capable of receiving a hash of features, and returning a boolean response predicting a citation, we can integrate the TACOS application with that service.

This achieves this integration via a few steps:

1. Feature extraction is handled by the legacy citation detector, by adding a fourth instance variable, `@features`.
2. We add a new, parallel detector to the application, `Detector::MlCitation`, which takes those features, assembles a payload, and communicates with the lambda. It then processes the response, and registers positive results as detections. This detector is guarded by a defensive clause, `.expected_env?`, which confirms that three environment variables are populated prior to taking any real actions.
3. The term model includes a consultation of the new detector, leveraging that guard clause so that we can deploy this code without the lambda yet being deployed.

One thing to note about this feature is that the needed env vars are described as optional, and kept out of `.env.test`. Adding them will require that a large number of tests that don't relate to this feature will now need cassettes, and a large number of tests in the future will likewise need cassettes unrelated to the feature under test. Instead, this defines a helper function that allows only relevant tests to define those env vars. We're using the helper clause pattern rather than a setup method because a few different test suites will occasionally need this feature - so this minimizes code repetition.

I'm not sure whether this is relevant, but I also want to note that the external ml_citation detector currently _always_ returns True, because it doesn't yet have the algorithm implemented. As a result, there's a test added here that is skipped because it will rely on the lambda being able to return either true or false.

Extra questions:
1. If it helps, the approach adopted here was not the first pattern I considered. An alternative approach, using a lookup class rather than a detector class, can be found in #245 .
2. This PR adds a new detector, leaving the legacy citation detector in place. I'm not sure this is needed - we build the DETECTOR_VERSION env var into the application to enable significant changes to detector behavior, which this change arguably fulfills. The change in this PR is a more incremental step toward that way of working, so I'm happy to propose this small step that still leaves the opportunity to move this new detector into the existing `Citation` record at a later step.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-183

### Accessibility

### Documentation

### ENV

- [x] All new ENV is documented in README.
- [ ] ENV has not changed.

### Stakeholders

NO dependencies are updated

NO migrations are included

### Code

added technical debt.

### Documentation

### Testing

### DIFF
--- a/app/models/detector/ml_citation.rb
+++ b/app/models/detector/ml_citation.rb
@@ -54,14 +54,41 @@ class Detector
       nil
     end
 
+    # lambda_path reads and returns the value of one environment variable.
+    #
+    # @note This is a public class method because the entire class ends up getting called in both class and instance
+    #   contexts, due to how detectors are built. The ideal state would be a private method, but that would require
+    #   changing how the class calls itself via the fetch method.
+    #
+    # @see Detector::MlCitation.expected_env?
+    # @see Detector::MlCitation.fetch
+    # @return String or nil
     def self.lambda_path
       ENV.fetch('DETECTOR_LAMBDA_PATH', nil)
     end
 
+    # lambda_secret reads and returns the value of one environment variable.
+    #
+    # @note This is a public class method because the entire class ends up getting called in both class and instance
+    #   contexts, due to how detectors are built. The ideal state would be a private method, but that would require
+    #   changing how the class calls itself via the fetch method.
+    #
+    # @see Detector::MlCitation.expected_env?
+    # @see Detector::MlCitation.fetch
+    # @return String or nil
     def self.lambda_secret
       ENV.fetch('DETECTOR_LAMBDA_CHALLENGE_SECRET', nil)
     end
 
+    # lambda_url reads and returns the value of one environment variable.
+    #
+    # @note This is a public class method because the entire class ends up getting called in both class and instance
+    #   contexts, due to how detectors are built. The ideal state would be a private method, but that would require
+    #   changing how the class calls itself via the fetch method.
+    #
+    # @see Detector::MlCitation.expected_env?
+    # @see Detector::MlCitation.fetch
+    # @return String or nil
     def self.lambda_url
       ENV.fetch('DETECTOR_LAMBDA_URL', nil)
     end


app/models/detector/citation.rb
```diff
@@ -135,13 +138,25 @@ def commas(phrase)
...
+    end
+
     # This builds one of the two main components of the Citation detector - the subpattern report. It uses each of the
     # regular expressions in the CITATION_PATTERNS constant, extracting all matches using the scan method.
     #
     # @return hash
     def pattern_checker(phrase)
       CITATION_PATTERNS.each_pair do |type, pattern|
-        @subpatterns[type.to_sym] = scan(pattern, phrase) if scan(pattern, phrase).present?
+        @subpatterns[type.to_sym] = scan(pattern, phrase)
```
JPrevost, line 159: Did you find the previous conditional did nothing useful here? I have no concerns specifically with removing the conditional but I'm curious if you realized it was never needed or if something changed to no longer need it.

app/models/detector/ml_citation.rb
```diff
@@ -0,0 +1,126 @@
...
+    #
+    #   @param phrase String. Often a `Term.phrase`.
+    #   @return Nothing intentional. Data is written to Hash `@detections` during processing.
+    def initialize(phrase)
+      return unless self.class.expected_env?
+
+      external_data = fetch(phrase)
+      return if external_data == 'Error'
+
+      @detections = external_data
```
JPrevost, line 17: This was a bit confusing to me as I found myself scanning for a method `external_data` rather than looking above for the variable.

Would you consider something like:
```ruby
@detections = fetch(phrase)
return if external_data == 'Error'
```

app/models/detector/ml_citation.rb
```diff
@@ -0,0 +1,126 @@
...
+    def self.lambda_path
+      ENV.fetch('DETECTOR_LAMBDA_PATH', nil)
+    end
+
+    def self.lambda_secret
+      ENV.fetch('DETECTOR_LAMBDA_CHALLENGE_SECRET', nil)
+    end
+
+    def self.lambda_url
+      ENV.fetch('DETECTOR_LAMBDA_URL', nil)
```
JPrevost, line 68: Do these three lambda methods get called anywhere outside this method? They feel like that should move to the private methods if not.

app/models/detector/ml_citation.rb
```diff
@@ -0,0 +1,126 @@
...
+    def fetch(phrase)
+      lambda = define_lambda
+      payload = define_payload(phrase)
+
+      response = lambda.post(self.class.lambda_path, payload.to_json)
+
+      if response.status == 200
+        JSON.parse(response.body)['response'] == 'true'
+      else
+        Rails.logger.error(response.body)
```
JPrevost, line 119: I don't recall if `Rails.logger.error` automatically triggers a Sentry alert. If not, I'd encourage an explicit Sentry error log here so we can understand if/how frequently this happens.

test/test_helper.rb
```diff
@@ -100,3 +101,22 @@ def auth_teardown
     end
   end
 end
+
```
JPrevost, line 104: I wonder if we should leave this helper in the `ml_citation_test` file for now until it is needed elsewhere. It seems unlikely that we'd actually need this helper anywhere else so it feels a bit more complex to have to look here rather than just having it at the top of the file being tested? This is not a strong feeling so if you feel it is better here that works for me.

JPrevost: Oh, I see. You use it it both ml_citation_test and term_test already. I'm leaving my comment so you can see what I was thinking... however, I now think this is the correct location for this helper.

test/models/detector/ml_citation_test.rb
```diff
@@ -0,0 +1,162 @@
...
+    test 'expected_env? returns false without DETECTOR_LAMBDA_URL set' do
+      with_enabled_mlcitation do
+        assert_predicate(Detector::MlCitation, :expected_env?)
+        ClimateControl.modify DETECTOR_LAMBDA_URL: nil do
+          assert_equal false, Detector::MlCitation.expected_env?
+        end
+      end
+    end
+
+    # These next three tests may not be needed, since the expected_env? method is tested above.
```
JPrevost, line 46: This comment feels a bit fragile as someone could insert a test in between any of these at some point. Maybe either drop the tests you feel are redundant or add the a more focused comment to each test such as:
`This test may not be needed, since the expected_env? method is also tested explicitly`

test/models/detector/ml_citation_test.rb
```diff
@@ -0,0 +1,162 @@
...
+          prediction = Detector::MlCitation.new('ping')
+
+          assert_instance_of Detector::MlCitation, prediction
+
+          assert(prediction.detections)
+        end
+      end
+    end
+
+    test 'lookup returns nil when challenge_secret is wrong' do
```
JPrevost, line 99: This test is doing a lot of work by handling "all" of the possible error states. I'm not suggesting we test _every_ possible error that could come back, but I think a test that has a name that is clear it is testing non-200 status would be great.

```ruby
test 'non 200 http status responses result in no detection' do
...
end
```

JPrevost (changes requested):
This looks good. It's a bit hard to test without something to call locally but I think I'm okay with that for now and I'll spend more time on the integration aspects when the lambda is a bit further along.

Not all of my comments require change, but I suspect something in them will drive you to want to make small tweaks. If not, let me know and I'll switch the status to approved.

app/models/detector/citation.rb
```diff
@@ -67,10 +67,13 @@ def detection?
     # @return Nothing intentional. Data is written to Hashes `@subpatterns`, `@summary`,
     #   and `@score` during processing.
     def initialize(phrase)
+      @features = {}
       @subpatterns = {}
       @summary = {}
       pattern_checker(phrase)
       summarize(phrase)
+      extract_features
+      @subpatterns.delete_if { |_, v| v == [] }
```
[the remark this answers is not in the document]
matt-bernhardt: This cleanup is needed because we're dropping the conditional in on line 144 (the next comment). The current sequence only populates `@subpatterns` for cases where there's a match:

`{lastnames: ["Hello,"]}`

However, for the `@features` variable we need those unmatched cases to be zeros.

With this PR, the initial state of `@subpatterns` would be

`{apa_volume_issue: [], no: [], pages: [], pp: [], vol: [], year_parens: [], brackets: [], lastnames: ["Hello,"], quotes: []}`

In that state, the `extract_features` command can very easily convert that to
`{apa_volume_issue: 0, no: 0, pages: 0, pp: 0, vol: 0, year_parens: 0, brackets: 0, lastnames: 1, quotes: 0}`

(and then merge that with the `@summary` variable for the full set of features).

This line then completes the cleanup of `@subpatterns`, and gets it back to just

`{lastnames: ["Hello,"]}`

app/models/detector/ml_citation.rb
matt-bernhardt: They don't. I wanted them to be private methods, but was having problems getting them referenced as private methods in some cases. the `.expected_env?` method that references these (among other references in this model) gets called both as instance and class methods, and making this public was the only way I could make sure that they were reliably available in all cases.

I'm not particularly happy about this arrangement, so I'm happy to see if I missed something.

app/models/detector/ml_citation.rb
matt-bernhardt: Yeah, good point - I'll add that.

app/models/detector/citation.rb
```diff
@@ -135,13 +138,25 @@ def commas(phrase)
       phrase.count(',')
     end
 
+    # This converts the already-built @subpatterns and @summary instance variables into the @features instance variable,
+    # which has a format suitable for sending to our prediction algorithm.
+    def extract_features
```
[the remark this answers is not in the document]
matt-bernhardt: Yeah, I think this is an option to consider, but agree with you that it should be delayed a bit. I'd be very interested in exploring the refactor to shift the feature extraction a bit, and considering this option in that context.

app/models/detector/ml_citation.rb
matt-bernhardt: I'm using this comment thread as a bit of code therapy, but may end up saying some of this in the ticket if it feels like it'd be useful long-term.

That specific arrangement doesn't work because `external_data` never gets populated, so the return clause can't do anything. I could change it to `return if @detections == 'Error'`, but... without anything else in the method I'm not sure what that would accomplish. The entire method could just be

```ruby
def initialize(phrase)
  return unless self.class.expected_env?
  
  @detections = fetch(phrase)
end
```

That has the advantage of being nice and simple, so maybe that's the winner. What I'm proposing initially, though, tries to prevent errors from being passed upward - hence the early return prior to `@detections` being populated in case something goes wrong getting a response back from the lambda. I'm not sure that the Term model or anything farther up the application needs to know that the communication with the citation detector went haywire?

The categorization process, the record-keeping in the term model, and the GraphQL response can all do their work without any input from this detector, so my thinking initially was that this detector either communicates a result back, or it handles errors on its own (with the coming addition of sending them to Sentry).

If that makes sense to you, I'm open to trying something like this?

```ruby
    def initialize(phrase)
      return unless self.class.expected_env?

      return if fetch(phrase) == 'Error'

      @detections = fetch(phrase)
    end
```

(This may yet need some work, as trying it locally causes three tests to fail with cassette problems. But it should be manageable)

matt-bernhardt: Ah... I think I see why the cassette problems happen. I was anticipating maybe there being some caching magic in place that would prevent multiple calls to `fetch(phrase)` from actually hitting the lambda twice - but that doesn't seem to be the case.

That leaves me with wanting:
1. Hit the lambda once, and check the result.
2. Return quickly, changing nothing, if that result is anything other than a valid response
3. Populate the `@detections` instance variable if the result _is_ a valid response

Ooh... I think I've got it. I'll add the new option to my next commit.

app/models/detector/ml_citation.rb
matt-bernhardt: Ok, the TL;DR is that it turns out they need to be public. I've tried making them private, and tried making them protected. Just moving them below the `private` declaration seems like it works, but CodeClimate and Rubocop point out that this isn't the way to make private class methods. Using the actual private_class_method prefix makes the linter happy, but then tests fail galore.

Ditto with trying a protected method.

The issue seems to boil down to the variety of ways that this method gets called. Externally, everything gets called as class methods - but then within the record method, the class dog-foods itself via instantiation, and the initialize method in turn calls expected_env? - which means that method gets used in both contexts, which means that the methods it calls being private invites a world of pain and suffering.

The sound you hear is me forcing myself not to just refactor all of our detector classes to avoid this dual-context situation. It probably needs to get added to our list of tech debt, but for right now having these be public methods is the way to go.

matt-bernhardt:
Ok, I think I've gotten the changes made that feel feasible without picking at some very dangerous threads - and it's ready for re-review on Monday.

app/models/detector/citation.rb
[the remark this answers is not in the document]
matt-bernhardt: I think this is a middle-ground state, and should be simplified in the near future. While the detector lambda isn't actually in service because we're working through the deploy process (and the ticket to implement the prediction capability isn't complete), we need to preserve the legacy citation detector.

Once the new system is fully in place, we will be free to refactor the feature extraction to not use this type of approach.

PR #361, opened 2026-08-06
Update to Rails 8.1

matt-bernhardt:
This incrementally upgrades the application from 7.2 to 8.0 and then on to 8.1, following the general pattern I see from Isra and Jeremy doing this for the other Rails apps.

### Ticket(s)

https://mitlibraries.atlassian.net/browse/TCO-216


config/environments/development.rb
```diff
@@ -17,52 +15,48 @@
...
-    config.public_file_server.headers = { "Cache-Control" => "public, max-age=#{2.days.to_i}" }
+    config.public_file_server.headers = { "cache-control" => "public, max-age=#{2.days.to_i}" }
   else
     config.action_controller.perform_caching = false
-
-    config.cache_store = :null_store
   end
 
+  # Change to :null_store to avoid any caching.
+  config.cache_store = :memory_store
```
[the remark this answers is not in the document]
matt-bernhardt: This feedback is at odds with the Rails documentation on caching in development: https://guides.rubyonrails.org/caching_with_rails.html#caching-in-development

config/environments/production.rb
```diff
@@ -6,47 +6,32 @@
...
 
-  # Ensures that a master key has been made available in ENV["RAILS_MASTER_KEY"], config/master.key, or an environment
-  # key such as config/credentials/production.key. This key is used to decrypt credentials (and other encrypted files).
-  # config.require_master_key = true
+  # Turn on fragment caching in view templates.
+  config.action_controller.perform_caching = true
 
   # Disable serving static files from `public/`, relying on NGINX/Apache to do so instead.
-  config.public_file_server.enabled = true
+  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
```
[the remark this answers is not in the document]
matt-bernhardt: After looking at our Heroku config and reasoning through this feature a bit more, I'm walking back this change. the public file server will be set to always be enabled in environments which use the production config.

public/404.html
```diff
@@ -1,67 +1,135 @@
...
+        font-size: 75%;
+      }
+
+      main article br {
+        display: none;
+
+        @media(min-width: 48em) {
+          display: inline;
+        }
+      }
```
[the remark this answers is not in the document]
matt-bernhardt: This comment applies to all the feedback recieved about these `public/*.html` templates - I don't think Copilot is correct in flagging this:

From: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Nesting/At-rules

> Any [at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Syntax/At-rules) whose body contains style rules can be nested inside another style rule using CSS nesting. Style rules nested inside at-rules take their nesting selector definition from the nearest ancestor style rule.

That article specifically includes examples of `@media` queries being nested inside selector rules.

Additionally, inspecting these pages as served in the review app, I see this rule taking effect at different page sizes on the 406-unsupported-browser example page.

config/environments/production.rb
```diff
@@ -6,47 +6,32 @@
...
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.asset_host = "http://assets.example.com"
 
-  # Specifies the header that your server uses for sending files.
-  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
-  # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
-
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
```
jazairi, line 31: Kinda weird that we've always had this set to local, but I don't see any need to change it since we don't store files in TACOS. 🤷‍♀️

config/ci.rb
```diff
@@ -0,0 +1,22 @@
+# Run using bin/ci
+
+CI.run do
+  step "Setup", "bin/setup --skip-server"
+
+  step "Style: Ruby", "bin/rubocop"
+
+  step "Security: Importmap vulnerability audit", "bin/importmap audit"
```
jazairi: This can be commented out, since we're not using importmaps. (I think? We don't in our other repos, at least.)

config/puma.rb
```diff
@@ -37,6 +42,9 @@
   Barnes.start # Must have enabled worker mode for this to block to be called
 end
 
+# Run the Solid Queue supervisor inside of Puma for single-server deployments.
+# plugin :solid_queue if ENV["SOLID_QUEUE_IN_PUMA"]
```
jazairi: Fwiw, we have this line uncommented in our other repos, but it's probably okay to leave it commented out since I don't think we're actually using it yet.

jazairi (changes requested):
I'd suggest including the Rails health check route ([as in here](https://github.com/MITLibraries/thing/blob/main/config/routes.rb#L2-L4)). Other suggestions are non-blocking.

config/ci.rb
matt-bernhardt: I noticed that this had been flagged in one of the other Rails 8 upgrades, and wasn't sure whether to disable or not I'm happy to take it out in the name of consistency - I agree that we don't use them in this app, at least.

config/environments/production.rb
matt-bernhardt: Yeah, if we were actually going to use this, it would need to be some other value (like ETD is). It looks like TIMDEX leaves out this directive entirely, while TIMDEX-UI also has it set to local.

I'm not opposed to having a coordinated value for these, but I think it'd be best to tackle that specifically, and not here.

matt-bernhardt:
Hey @jazairi - I made a number of the changes you mentioned. I got confused when adding the health check route broke a bunch of tests, but then noticed that it had already been added about halfway down the routes file (I swear I'd added it during the upgrade at one point, but the blame report for this file indicates that it's been here for a while).

Hopefully re-organizing the routes file to put it at the top will help with consistency and not confuse future-me so much.

jazairi:
@matt-bernhardt Ah, that explains why the routes file wasn't touched. Sorry, should've checked the current state!

matt-bernhardt:
No worries! I saw your requested change, and thought "wait, I thought I did add that" - and then saw that it wasn't in the diff, and thought maybe I slow-walked it for some reason. It wasn't until I then added it and saw things break that I realized it was already there, and had been for some time.

So, I definitely missed that it was already there too. GMTA :-)

